PR #41, opened 2022-12-23
Change HF training script to use input splits

polm:
This changes the HF training script to use separate `train.spacy` and `dev.spacy` files, just like normal spaCy training. The README has also been updated to cover the new training process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,22 +135,29 @@ python -m prodigy db-out my_ner_data  > ner_data.jsonl
 The format of the exported annotations contains all the data you need to train a smaller model downstream. Each example
 in the dataset contains the original text, the tokens, span annotations denoting the entities, etc.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -146,11 +146,11 @@ We've also included an experimental script to load in the `.spacy` binary format
 
 ```
 pip install -r requirements_train.txt
-python -m prodigy data-to-spacy ./data/annotations2 --ner my_ner_data -es 0
-python scripts/train_hf_ner.py ./data/annotations2/train.spacy ner-model
+python -m prodigy data-to-spacy ./data/annotations2 --ner my_ner_data
+python scripts/train_hf_ner.py ./data/annotations2/train.spacy ./data/annotations2/dev.spacy ner-model
```
svlandeg, line 150: The biggest advantage of splitting out beforehand is that we can show the three steps: `data-to-spacy`, then either `spacy train` or `train_hf_ner.py`, or both. So I wouldn't repeat the `data-to-spacy` command here, and it allows us to get rid of this ugly `annotations2` dir.

At L138, I'd rewrite slightly so that the `bash` code bit gets split into two, one for `data-to-spacy` and one for `spacy-train`, making it clear to readers how they can combine them.

svlandeg: Also I'd like to keep the `-es` parameter in the `data-to-spacy` command as users often will want to control the % of the split.

svlandeg (approved):
Thanks, looks good to me! 

I'll give this a spin later today, but will already merge this in so we can test everything further.

PR #42, opened 2022-12-26
Add Azure pipeline

svlandeg:


azure-pipelines.yml
```diff
@@ -0,0 +1,66 @@
...
+
+  - script: |
+      python -m pip install -U build pip setuptools
+      pip install -U -r requirements.txt
+    displayName: 'Install dependencies'
+
+  - task: DeleteFiles@1
+    inputs:
+      contents: 'openai-prodigy-recipes'
+    displayName: 'Delete source directory'
```
honnibal, line 63: What's with this part?

honnibal:
Yay thanks!

I was actually just starting on the A/B recipe, so this is timely :)

azure-pipelines.yml
svlandeg: Oh we don't need this bit if we're not building from sdist, I'll remove it.

PR #43, opened 2022-12-26
Get tests running

svlandeg:
WIP

- remove Python 3.6
- use secret env var to install Prodigy


svlandeg:
I'll still be following up to improve on this, but it seems to work somewhat for now, and it'll be easier to continue experimentation when it's merged into `main`

PR #45, opened 2022-12-26
Prompt A/B evaluation recipe

honnibal:
Add a recipe for A/B evaluation of prompts.

[AB evaluation]( https://prodi.gy/features/ab-evaluation) is basically like a blind “taste test”. You line up two functions to compare and a sequence of inputs, and then answer a sequence of blind `compare(functionA(input[i]), functionB(input[i]))` questions. To extend the “taste test” analogy a bit further, it’s like evaluating two ingredients in a sequence of dishes, like this: https://youtu.be/PwM0AeB6N8o?t=750 . Instead of parmesan cheeses, we’ll be evaluating prompt templates, and instead of dishes (like "raw", "in alfredo sauce", "on pizza"), we’ll be using a sequence of template arguments. The results the annotator will be ‘tasting’ are the OpenAI API responses to the rendered templates.

The goal of the A/B evaluation is to quickly let someone compare the quality of outputs from two prompts in a quantifiable and blind way. Other annotators can also go back over that data later and see whether they agree with the decisions. So it’s a methodologically sound way to approach things, even if the decisions are subjective.

I also have an idea for some bonus future cleverness, that maybe I'll only get to in a future PR. I've always wanted to extend the A/B evaluation in Prodigy to a multi-participant tournament. I could imagine a Bayesian approach to this: you start out with a flat prior over the candidates, and you determine which pair of candidates will give you the most information about which one is the highest rank. Querying a candidate that you have less information about can potentially produce a bigger update, but you also want to query more from the candidates which are ranking highly. This would let the user compare more than two prompts at once.


recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+            selected = eg.get("accept", [])
+            if not selected or len(selected) != 1 or eg["answer"] != "accept":
+                continue
+            counts[selected[0]] += 1
+        print("")
+        if not counts:
+            msg.warn("No answers found", exits=0)
+        msg.divider("Evaluation results", icon="emoji")
+        pref, _ = counts.most_common(1)[0]
+        msg.good(f"You preferred {pref}")
```
svlandeg, line 115: This should guard against equal preference, I got:
```
=========================== ✨  Evaluation results ===========================
✔ You preferred prompt1.jinja2

prompt1.jinja2   2
prompt2.jinja2   2

```

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+import sys
+import time
+from pathlib import Path
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar, cast
+
+import httpx
+import jinja2
+import prodigy
+import prodigy.components.db
+import prodigy.components.preprocess
```
svlandeg, line 25:
```suggestion
```
Imports seem to be unnecessary

svlandeg:
With the `choice` interface, I suppose it isn't straightforward to show the options next to eachother instead of below eachother?

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+            for input_, responses in zip(input_batch, response_batch):
+                yield self._make_example(
+                    input_.id,
+                    self.display.render(**input_.prompt_args),
+                    responses,
+                    randomize=self.randomize,
+                )
+
+    def progress(self) -> float:
+        return 0.0
```
svlandeg, line 99: What does this do?

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+                continue
+            counts[selected[0]] += 1
+        print("")
+        if not counts:
+            msg.warn("No answers found", exits=0)
+        msg.divider("Evaluation results", icon="emoji")
+        pref, _ = counts.most_common(1)[0]
+        msg.good(f"You preferred {pref}")
+        rows = [(name, count) for name, count in counts.most_common()]
+        msg.table(rows, aligns=("l", "r"))
```
svlandeg, line 117: Ideally this would also show the template that got 0 responses, currently it's:
```
=========================== ✨  Evaluation results ===========================
✔ You preferred prompt2.jinja2

prompt2.jinja2   3
```

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+            ),
+            n=self.openai_n,
+            timeout_s=self.openai_timeout_s,
+        )
+        r.raise_for_status()
+        responses = r.json()
+        return [responses["choices"][i]["text"] for i in range(len(prompts))]
+
+    def _make_example(
+        self, id: str, input: str, responses: Dict[str, str], randomize: bool
```
svlandeg, line 168: Should we avoid using `id` and `input` as var names, as they clash with builtins?

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+@prodigy.recipe(
+    "ab.openai.prompts",
+    dataset=("Dataset to save answers to", "positional", None, str),
+    inputs_path=("Path to jsonl inputs", "positional", None, Path),
+    display_template_path=(
+        "Template for summarizing the arguments",
+        "positional",
+        None,
+        Path,
+    ),
```
svlandeg, line 195: I worry that this parameter will be confusing to users. Using it as a descriptive title means it'll always be pretty similar to one of the actual prompts, adding to the confusion.
I wonder whether instead we should have a more structured title on the Prodigy UI? So instead of 
```
Poem about just vibing out in the style of John Donne
```
it would read something more "technical", showing the actual slots literally, like 
```
Topic: "just vibing out"
Author: "John Donne"
```

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+        rows = [(name, count) for name, count in counts.most_common()]
+        msg.table(rows, aligns=("l", "r"))
+
+    def _get_response_batch(self, inputs: List[PromptInput]) -> List[Dict[str, str]]:
+        name1, name2 = self._choose_rivals()
+        prompts = []
+        for input_ in inputs:
+            prompts.append(self._get_prompt(name1, input_.prompt_args))
+            prompts.append(self._get_prompt(name2, input_.prompt_args))
+        responses = self._get_responses(prompts)
```
svlandeg, line 125: It might be nice to show the actual queries on the CLI when the `verbose` flag is used:
```suggestion
        if self.verbose:
            for prompt in prompts:
                rich.print(Panel(prompt, title="Prompt to OpenAI"))
        responses = self._get_responses(prompts)
```
(requires adding the correct imports, too)

recipes/openai_ab.py
```diff
@@ -0,0 +1,365 @@
...
+    )
+    return {
+        "dataset": dataset,
+        "view_id": "choice",
+        "stream": stream,
+        "on_exit": stream.on_exit,
+        "config": {
+            "batch_size": batch_size,
+            "exclude_by": "input",
+            "show_flag": True,
```
svlandeg, line 258: Do we need the flag for this recipe?

koaning:
I had a look at the PR and figured I might push two minor improvements. 

## Change 1: style

This is before: 

CleanShot 2023-01-26 at 12 19 16

This is after:

CleanShot 2023-01-26 at 12 31 15

It turned out that the whitespace that combes back from OpenAI actually gets rendered. So I called `.strip()` on that and that makes the text appear nice and centered. I also adapted the line width to be slightly smaller to accomodate for longer pieces of text. 

## Change 2: repeat 

Given that OpenAI has entropy, it seems to make sense to allow the user to judge the output of a prompt more than once. So I added a `--repeat` flag that allows the user to specify the number of times a prompt should be repeated. It's a small change, but it makes it much easier for me to get more data. With just five examples, I'm able to set `--repeat 3` and eventually get more data to base my prompt decision on. 

```
=========================== ✨  Evaluation results ===========================
✔ You preferred prompt1.jinja2

prompt1.jinja2   13
prompt2.jinja2    7
```

recipes/openai_ab.py
koaning: Hey yeah, it's not used anywhere. Removing for now.

recipes/openai_ab.py
koaning: Well spotted! Just wrote some code to handle this edge case.

recipes/openai_ab.py
koaning: It could just be passed as `"meta"` information on the example. I'll have a look!

recipes/openai_ab.py
koaning: Fixed! 

```
=========================== ✨  Evaluation results ===========================
✔ It's a draw!

prompt1.jinja2   0
prompt2.jinja2   0
```

koaning:
> With the choice interface, I suppose it isn't straightforward to show the options next to eachother instead of below eachother?

We could override the CSS settings, but in general we want this space because we typically generate longer sequences of text.

recipes/openai_ab.py
koaning: I just expanded `._make_example` to also include the meta information, which looks like this in the annotation interface: 

CleanShot 2023-01-26 at 13 52 07

recipes/openai_ab.py
koaning: I agree. Calling them `input_str` and `id_str` for now.

recipes/openai_ab.py
koaning: Apon reflection, I think I like it better with the header around. 

CleanShot 2023-01-26 at 14 02 27

It might still be nice to have it optional, but I'm not super bullish on that. I'll keep Matt's implementation here.

koaning:
I've incorporated Sofie's comments now. These were mostly small changes to tackle edge-cases but I did end up adding the meta information to the interface. It's a small detail, but it does make it clear what parameters were used to generate the prompt to OpenAI. 

CleanShot 2023-01-26 at 14 07 49

@honnibal I've used it a bunch locally and I like the recipe. I can move on to writing the docs for it, but I want to discuss a few things from my end. 

1. Does it make sense to allow the user to write the prompt from the command line, or do we strictly want to stick to template files here?
2. Technically, the OpenAI hyperparams might matter here. We currently don't allow these to change from the command line and we also don't add this information to the output. To keep things simple we probably don't want to include them, but it's worth a ponder. 
3. Is there a reason to format the example data as `{id: int, prompt_args: {**kwargs}}`? I don't know what the user might use the `id` for and I wonder if it might be simpler to just have the user supply `{"topic": "star wars"}` instead of `{"id": 0, "prompt_args": {"topic": "star wars"}}`. I may be overlooking something here though.

magdaaniol:
Should we refactor it to use the OpenAISuggester class?

koaning:
I think the reason why this uses `prompt_args` is related to Pydantic. It's a bit verbose, but it also seems fine and not a reason to do a big refactor now. I can also imagine that the `id` might be used to group items later.

koaning:
Adding a screenshot so I may use it in the README. 

CleanShot 2023-02-03 at 13 47 36

recipes/openai_ab.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+@prodigy.recipe(
+    # fmt: off
+    "ab.openai.prompts",
+    dataset=("Dataset to save answers to", "positional", None, str),
+    inputs_path=("Path to jsonl inputs", "positional", None, Path),
+    display_template_path=("Template for summarizing the arguments","positional", None, Path),
+    prompt1_template_path=("The first jinja2 prompt template","positional", None, Path),
+    prompt2_template_path=("Path to second jinja2 prompt template","positional", None, Path),
```
magdaaniol, line 203:
```suggestion
    prompt2_template_path=("Path to the second jinja2 prompt template","positional", None, Path),
```

README.md
```diff
@@ -503,6 +503,114 @@ OpenAI has a hard limit on the prompt size. You cannot have a prompt larger than
...
+python -m prodigy ab.openai.prompts dataset inputs_path display_template_path prompt1_template_path prompt2_template_path [--options] -F ./recipes/openai_ab.py
+```
+
+|        Argument        | Type  | Description                                          | Default              |
+|:----------------------:|-------|------------------------------------------------------|----------------------|
+| `dataset`              | str   | Prodigy dataset to save answers into                 |                      |
+| `inputs_path`          | Path  | Path to jsonl inputs                                 |                      |
+| `display_template_path`| Path  | Template for summarizing the arguments               |                      |
+| `prompt1_template_path`| Path  | The first jinja2 prompt template                     |                      |
+| `prompt2_template_path`| Path  | Path to second jinja2 prompt template                |                      |
```
magdaaniol, line 523:
```suggestion
| `prompt2_template_path`| Path  | Path to the second jinja2 prompt template                |                      |
```

koaning:
@magdaaniol, those were all good edits.

Good to merge?

honnibal:
Thanks a lot for all the improvements here! All is good with me.

magdaaniol (approved):

PR #48, opened 2022-12-27
Further cleanup of the HF training script

svlandeg:
Functionality changes:
- <s>Require passing in the `nlp` config file</s> Use a `Vocab` instance instead of hardcoding `spacy.blank("en")`
- Make the output dir optional in case users just want to run the training without saving the model

Cosmetic changes:
- Adding descriptions of the various parameters for a more verbose `--help` function on the CLI
- Added type annotations throughout the code
- Renamed some variables for easier code readability
- Rounding the performance numbers to 3 decimals makes the output more readable
- Suppressing the "zero division" warning
- `black` & `isort` formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ We've also included an experimental script to load in the `.spacy` binary format
 ```bash
 # First you need to install the HuggingFace library and requirements
 pip install -r requirements_train.txt
-python scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotations/dev.spacy -o hf-ner-model
+python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotations/dev.spacy -o hf-ner-model
 ```
 
 The resulting model will be saved to the `hf-ner-model/` directory.
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ We've also included an experimental script to load in the `.spacy` binary format
 ```bash
 # First you need to install the HuggingFace library and requirements
 pip install -r requirements_train.txt
-python scripts/train_hf_ner.py ./data/annotations/config.cfg ./data/annotations/train.spacy ./data/annotations/dev.spacy hf-ner-model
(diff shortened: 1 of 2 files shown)


polm:
Thanks for the cleanup. This all looks fine and I was able to successfully run training with the modifications. 

About requiring a config, I think this script would be better without that change. The spaCy pipeline is just used for the Vocab for deserializing the DocBin, so nothing in the input config makes a difference. In this particular case Prodigy already provides a config, so it's not a big deal to provide one, but as a user I would be confused about why I was being asked for a config here.

svlandeg:
> About requiring a config, I think this script would be better without that change. The spaCy pipeline is just used for the Vocab for deserializing the DocBin, so nothing in the input config makes a difference. In this particular case Prodigy already provides a config, so it's not a big deal to provide one, but as a user I would be confused about why I was being asked for a config here.

I've been doubting this change myself, it's just that it feels so weird/wrong to hardcode `"en"` like that. And as we're already making people use `.spacy` files, they will probably always run the `data-to-spacy` command so yea the `.config` will always be just sitting there. 

How about we just provide the directory path of the data generated by `data-to-spacy`? Then the fact that we use the config would be an implementation detail that users don't need to worry about. It does mean that the HF training command looks quite a bit different than the `spacy train` command - adding the config had the effect that they became pretty similar which I thought was nice. I can see it either way though.

polm:
Fair point about hard coding "en". We could use "xx" or something instead. Also, I had to test it to be sure it'd work, but it looks like we can just create a Vocab object without a pipeline and that works too:

```
from spacy.vocab import Vocab
voc = Vocab()
for doc in db.get_docs(voc):
    ...
```

So that would be another way to avoid asking for a config.

svlandeg:
>  it looks like we can just create a Vocab object without a pipeline and that works too:

Maybe that makes most sense, as we don't actually need the `nlp` object! Do you have time to make this change? If not I can do it later today.

polm:
Got it, I went ahead and made the changes, so I think that should be clean now.

PR #49, opened 2022-12-29
Add evaluation recipe for zero-shot

honnibal:
~~This script takes the corrected annotations as a gold-standard, and asks how well the zero- or few-shot response from OpenAI scores against that.~~

~~The code duplication here is definitely getting unfortunate, but I was reluctant to import from the recipe. Actually this is probably better as a recipe within the NER file? Then it can read the dataset directly from the DB.~~

Edit: Okay yeah went ahead and made it a recipe. I think this is better. I've left the script in there for discussion, assuming we like the recipe I'll remove the script.


recipes/openai_ner.py
```diff
@@ -256,7 +248,8 @@ def _get_ner_response(self, prompts: List[str]) -> List[str]:
         responses = r.json()
         return [responses["choices"][i]["text"] for i in range(len(prompts))]
 
-    def _parse_response(self, text: str) -> List[Tuple[str, List[str]]]:
+    @classmethod
+    def parse_response(cls, text: str) -> List[Tuple[str, List[str]]]:
```
honnibal, line 252: I made this a classmethod so I could use it as a helper without instantiating the class.

I didn't want to just make it a loose function, because the response parsing will depend on the template people use, so it's reasonable to expect people to want to subclass for this.

recipes/openai_ner.py
```diff
@@ -270,12 +263,29 @@ def _parse_response(self, text: str) -> List[Tuple[str, List[str]]]:
             if line and ":" in line:
                 label, phrases = line.split(":", 1)
                 label = _normalize_label(label)
-                if label in self.labels:
-                    if phrases.strip():
-                        phrases = [phrase.strip() for phrase in phrases.strip().split(",")]
-                        output.append((label, phrases))
+                if phrases.strip():
+                    phrases = [phrase.strip() for phrase in phrases.strip().split(",")]
+                    output.append((label, phrases))
```
honnibal, line 268: I removed the `if label in self.labels` check here, because we're doing the same filtering on that just below, and it saved me from passing the labels list in here too.

scripts/evaluate_openai.py
```diff
@@ -0,0 +1,159 @@
+"""Evaluate the accuracy of the OpenAI zero-shot NER suggestions,
+when compared to the corrected data.
+
+We want to make sure we use the same responses we used during
+annotation, so we read those in from the jsonl. We also read in
+the spaCy-formatted evaluation samples.
+"""
```
honnibal, line 7: We can remove this whole script if we prefer the recipe version.

PR #50, opened 2023-01-04
Add OpenAI Textcat recipes

ljvmiranda921:
## Description

This Pull Request does two things:

- Add two recipes for text classification with the help of ChatGPT
- Refactor the `OpenAISuggester` as a base class so that it's more reusable

## Add recipes for text classification

This PR adds two recipes, `textcat.openai.correct` and `textcat.openai.fetch`, to perform text classification in a streaming and batch manner. I also removed the  `textcat.openai.suggest` because the `fetch` command already fulfills the same purpose.

Unlike the previous state of this PR, I have now combined binary and multiclass textcat into a single recipe, where the logic is based on the number of labels (i.e., len(labels)>1 then multiclass). The template also follows a similar logic using if-else statements found in Jinja2. 

I've also made the `prompt_path` a positional argument because I think it reads better, i.e.:

> Help me annotate <input_path> using this <prompt_path>

Below are a few examples:

### Example 1: Binary classification

In this example, we perform binary classification by passing a single label to `--labels`. The `view_id` also changes with respect to this value.

```sh
python3 -m prodigy textcat.openai.correct textcat_test \
    data/reddit_r_cooking_sample.jsonl \ # input path
    templates/textcat_prompt.jinja2 \ # template path
    --labels recipe
    -F recipes/openai_textcat.py
    --verbose
```

### Example 2: Multiclass / Multilabel classification

For multiclass (exclusive) classification, we pass the parameter
`--exclusive-classes` or `-E` to state that we only want a single value. The
`config.choice` parameter changes accordingly:

```sh
python3 -m prodigy textcat.openai.correct textcat_test \
    data/reddit_r_cooking_sample.jsonl \ # input path
    templates/textcat_prompt.jinja2 \ # template path
    --labels recipe,feedback,question
    -F recipes/openai_textcat.py
    --exclusive-classes
    --verbose
```

For multilabel classification, we don't set the `-E` flag. Notice that it's now a checkbox instead of a choice button:

```sh
python3 -m prodigy textcat.openai.correct textcat_test \
    data/reddit_r_cooking_sample.jsonl \ # input path
    templates/textcat_prompt.jinja2 \ # template path
    --labels recipe,feedback,question
    -F recipes/openai_textcat.py
    --exclusive-classes
    --verbose
```

## Refactor OpenAISuggester

I went ahead and attempted to generalize the `OpenAISuggester` class. It is now located in `recipes/openai.py`. It is now an abstract base class (`abc.ABC`) and has an abstractmethod `parse_response`. This means that we need to inherit `OpenAISuggester` and implement the `parse_response(example: Dict, response: str)` function.

The role of `parse_response` is to convert whatever ChatGPT returns (`response`) and assign
the parsed values into appropriate keys in a copy of `example`. You can see an example implementation of this in `recipes:openai_textcat.py:TextCatOpenAISuggester`. 

## Changelog
- January 18: did a few refactors after initial review.


recipes/openai_textcat.py
```diff
@@ -0,0 +1,470 @@
...
+        verbose=verbose,
+        segment=segment,
+        openai_api_key=api_key,
+        openai_api_org=api_org,
+    )
+    for eg in examples:
+        openai.add_example(eg)
+    stream = list(srsly.read_jsonl(input_path))
+    stream = openai(tqdm.tqdm(stream), batch_size=batch_size, nlp=nlp)
+    stream = _prefer_gpt(stream, chatgpt_bias)
```
ljvmiranda921, line 317: I added a sorter here that allows users to bias examples where ChatGPT3 thinks a particular text is a recipe. The implementation for now is like:

```python
if chatgpt3_thinks_positive:
    probs = bias
else:
    probs = (1 - bias)
```

Then I feed it to the `prodigy.components.sorters.Probability` component. This allows us to affect the "stream."

I'm still figuring out what's the best way to "know" if the positive samples are rare based on ChatGPT3 predictions. I can do a count, but that means calling another recipe for the whole corpus with some parameters (then calling `textcat.manual` for correction).

recipes/openai_textcat.py
```diff
@@ -0,0 +1,470 @@
...
+                {"view_id": "classification"},
+                {"view_id": "html", "html_template": HTML_TEMPLATE},
+            ],
+            "show_flag": True,
+            "global_css": CSS_FILE_PATH.read_text(),
+        },
+    }
+
+
+def _prefer_gpt(stream, bias: float) -> Iterable[Dict]:
```
ljvmiranda921, line 338: My current implementation of the said sorter. The idea is for us to bias samples that ChatGPT thinks as positive.

ljvmiranda921 (commented):
Some initial implementations! Feel free to take a quick look!

koaning:
@ljvmiranda921 [the OpenAI help](https://help.openai.com/en/articles/5955604-how-can-i-solve-429-too-many-requests-errors) suggests using an exponential backoff for the rate limits. I guess right now this PR is using a linear variant, so that might help with the timeouts.

templates/textcat_prompt.jinja2
```diff
@@ -0,0 +1,11 @@
+From the text below, determine whether or not it contains a recipe for a
```
koaning, line 1: It seems that this template is hard-coded to be about "recipes", is that intentional? While we could require the users to make their own template for their use-case, it seems more friendly to have a general prompt where they only need to provide us with the labels that they are interested in. 

The [ner example](https://github.com/explosion/prodigy-openai-recipes/blob/main/templates/ner_prompt.jinja2) gives a nice example of a template automatically generated out of labels passed as input. Would such a template be preferable here as well?

koaning: I suppose this also opens up the question: is it the intent for this recipe to only support binary classification? That's certainly possible, just want to make sure it's the intent.

ljvmiranda921: Hi @koaning ! Thanks for the review! Yeah my original intention was to have it work on the recipe comments dataset (binary classification). But looking back I think it makes sense to have it cover non-exclusive classes. I'll update this PR!

koaning: Grand. Let me know if you appreciate any help or when you've come around to making the template general. I'd like to give this recipe a spin 👍

ljvmiranda921: Nice! I'm only getting to it last Friday but I set myself a hard deadline tomorrow (Tuesday). If I'm still stuck by that day, I'll ping!

ljvmiranda921: @koaning I think this is ready for a spin! I updated the README for some examples you can try out!

recipes/openai_textcat.py
```diff
@@ -0,0 +1,535 @@
...
+    ) -> Iterable[Dict]:
+        """Parse the examples in the stream and set up labels
+        to display in the Prodigy UI.
+        """
+        stream = prodigy.components.preprocess.add_tokens(nlp, stream, skip=True)  # type: ignore
+        for example in stream:
+            example = copy.deepcopy(example)
+            response = self._parse_response(example["openai"]["response"])
+            example["answer"] = response["answer"] == "accept"
+            example["meta"]["reason"] = response["reason"]
```
koaning, line 202: I tried running the base recipe and I think I hit a bug. I have a `doubt.jsonl` file 

```
python -m prodigy textcat.openai.suggest openai-text-demo doubt.jsonl excitement  -F recipes/openai_textcat.py
```

This resulted in this error. 

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/prodigy/__main__.py", line 61, in <module>
    controller = recipe(*args, use_plac=True)
  File "cython_src/prodigy/core.pyx", line 339, in prodigy.core.recipe.recipe_decorator.recipe_proxy
  File "cython_src/prodigy/core.pyx", line 366, in prodigy.core._components_to_ctrl
  File "cython_src/prodigy/core.pyx", line 125, in prodigy.core.Controller.__init__
  File "cython_src/prodigy/components/feeds.pyx", line 104, in prodigy.components.feeds.Feed.__init__
  File "cython_src/prodigy/components/feeds.pyx", line 150, in prodigy.components.feeds.Feed._init_stream
  File "cython_src/prodigy/components/stream.pyx", line 107, in prodigy.components.stream.Stream.__init__
  File "cython_src/prodigy/components/stream.pyx", line 58, in prodigy.components.stream.validate_stream
  File "recipes/openai_textcat.py", line 209, in filter_suggestions
    for example in stream:
  File "recipes/openai_textcat.py", line 202, in format_suggestions
    example["meta"]["reason"] = response["reason"]
KeyError: 'meta'
```

I was able to fix this by adding this to line 202:

```python
if "meta" not in example:
    example["meta"] = {}
```

recipes/openai_textcat.py
```diff
@@ -0,0 +1,535 @@
...
+
+    stream = list(srsly.read_jsonl(input_path))
+    stream = openai(
+        tqdm.tqdm(stream), batch_size=batch_size, nlp=nlp, negative_bias=negative_bias
+    )
+
+    return {
+        "dataset": dataset,
+        "view_id": "blocks",
+        "stream": stream,
```
koaning, line 342: When I ran this recipe: 

```
python -m prodigy textcat.openai.suggest openai-text-demo doubt-justtext.jsonl excitement  -F recipes/openai_textcat.py
```

I got a labelling interface that does not show the label. 

CleanShot 2023-01-17 at 12 11 00

The recipe uses a [classification interface](https://prodi.gy/docs/api-interfaces#classification) in the blocks, so this should be fixed by adding a `label` key to each item in the stream with the label name in it.

recipes/openai_textcat.py
```diff
@@ -0,0 +1,535 @@
...
+    examples_path: Optional[Path] = None,
+    prompt_path: Path = DEFAULT_PROMPT_PATH,
+    max_examples: int = 2,
+    verbose: bool = False,
+):
+    """Get bulk TextCat suggestions from an OpenAI API, using zero-shot or
+    few-shot learning.  The results can then be corrected using the
+    `textcat.manual` recipe.
+
+    Here, we use ChatGPT to suggest if a particular text talks about a recipe or
```
koaning, line 304: This recipe is general now, so instead of "recipe" maybe call it "label"?

recipes/openai_textcat.py
koaning: I may be missing some context because I wasn't around at the beginning of this recipe, but shouldn't we add an extra HTML block that makes it clear what OpenAI thought of the text. Having the reason around certainly helps, but I wonder if we can visualise the answer from OpenAI more explicitly. 

Maybe something more like below (made this in pixelmator real quick): 

CleanShot 2023-01-17 at 12 20 11

recipes/openai_textcat.py
```diff
@@ -0,0 +1,535 @@
...
+        return self.prompt_template.render(text=text, label=label)
+
+    def _get_textcat_response(self, prompts: List[str], delay: int = 0) -> List[str]:
+        headers = {
+            "Authorization": f"Bearer {self.openai_api_key}",
+            "OpenAI-Organization": self.openai_api_org,
+            "Content-Type": "application/json",
+        }
+        time.sleep(delay)
+        r = _retry429(
```
koaning, line 233: While running the recipe I also got this error: 

```
future: <Task finished name='Task-32' coro=<RequestResponseCycle.run_asgi() done, defined at /home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py:394> exception=HTTPStatusError("Server error '503 Service Unavailable' for url 'https://api.openai.com/v1/completions'\nFor more information check: https://httpstatuses.com/503")>
Traceback (most recent call last):
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 399, in run_asgi
    self.logger.error(msg, exc_info=exc)
  File "/usr/lib/python3.8/logging/__init__.py", line 1463, in error
    self._log(ERROR, msg, args, **kwargs)
  File "/usr/lib/python3.8/logging/__init__.py", line 1577, in _log
    self.handle(record)
  File "/usr/lib/python3.8/logging/__init__.py", line 1586, in handle
    if (not self.disabled) and self.filter(record):
  File "/usr/lib/python3.8/logging/__init__.py", line 807, in filter
    result = f.filter(record)
  File "cython_src/prodigy/util.pyx", line 145, in prodigy.util.ServerErrorFilter.filter
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 396, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
    await super().__call__(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/cors.py", line 86, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/cors.py", line 142, in simple_response
    await self.app(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 25, in __call__
    response = await self.dispatch_func(request, self.call_next)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/prodigy/app.py", line 205, in reset_db_middleware
    response = await call_next(request)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 45, in call_next
    task.result()
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 38, in coro
    await self.app(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/routing.py", line 580, in __call__
    await route.handle(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/routing.py", line 241, in handle
    await self.app(scope, receive, send)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/routing.py", line 52, in app
    response = await func(request)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/fastapi/routing.py", line 226, in app
    raw_response = await run_endpoint_function(
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/fastapi/routing.py", line 161, in run_endpoint_function
    return await run_in_threadpool(dependant.call, **values)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/starlette/concurrency.py", line 40, in run_in_threadpool
    return await loop.run_in_executor(None, func, *args)
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/prodigy/app.py", line 434, in get_session_questions
    return _shared_get_questions(req.session_id, excludes=req.excludes)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/prodigy/app.py", line 399, in _shared_get_questions
    tasks = controller.get_questions(session_id=session_id, excludes=excludes)
  File "cython_src/prodigy/core.pyx", line 221, in prodigy.core.Controller.get_questions
  File "cython_src/prodigy/core.pyx", line 222, in prodigy.core.Controller.get_questions
  File "cython_src/prodigy/components/feeds.pyx", line 379, in prodigy.components.feeds.Feed.get_batch
  File "cython_src/prodigy/components/feeds.pyx", line 323, in prodigy.components.feeds.Feed._enqueue_tasks
  File "cython_src/prodigy/components/stream.pyx", line 170, in prodigy.components.stream.Stream.__next__
  File "cython_src/prodigy/components/stream.pyx", line 174, in prodigy.components.stream.Stream.__next__
  File "cython_src/prodigy/components/filters.pyx", line 80, in filter_inputs
  File "recipes/openai_textcat.py", line 211, in filter_suggestions
    for example in stream:
  File "recipes/openai_textcat.py", line 198, in format_suggestions
    for example in stream:
  File "cython_src/prodigy/components/preprocess.pyx", line 164, in add_tokens
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/spacy/language.py", line 1539, in pipe
    for doc in docs:
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/spacy/language.py", line 1583, in pipe
    for doc in docs:
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/spacy/language.py", line 1580, in <genexpr>
    docs = (self._ensure_doc(text) for text in texts)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/spacy/language.py", line 1529, in <genexpr>
    docs_with_contexts = (
  File "cython_src/prodigy/components/preprocess.pyx", line 157, in genexpr
  File "recipes/openai_textcat.py", line 182, in stream_suggestions
    responses = self._get_textcat_response(prompts, delay=0)
  File "recipes/openai_textcat.py", line 249, in _get_textcat_response
    r.raise_for_status()
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/httpx/_models.py", line 736, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Server error '503 Service Unavailable' for url 'https://api.openai.com/v1/completions'
For more information check: https://httpstatuses.com/503
```

We might want to retry these as well?

recipes/openai_textcat.py
koaning: I noticed another issue that popped up with the missing label. Imagine that I'd be interested in detecting two emotions. 

```bash
# Excitement
python -m prodigy textcat.openai.suggest openai-text-demo doubt.jsonl excitement  -F recipes/openai_textcat.py

# Anger
python -m prodigy textcat.openai.suggest openai-text-demo doubt.jsonl excitement  -F recipes/openai_textcat.py
```

Then both of these calls should start at the beginning of my `doubt.jsonl` and not skip any examples as OpenAI tries to varify the emotion. However, if there is no label key on the example then the `_task_hash` on the example will be the same for both emotions.

recipes/openai_textcat.py
koaning: The same error pops up in the multilabel file.

recipes/openai_textcat_multilabel.py
```diff
@@ -0,0 +1,527 @@
+import copy
```
koaning, line 1: Is there a reason why this had to happen in a separate file? The Prodigy [`textcat.correct`](https://prodi.gy/docs/recipes/#textcat-correct) recipe will adapt based on value for `--labels`. Is there a reason that cannot happen here?

recipes/openai_textcat_multilabel.py
```diff
@@ -0,0 +1,527 @@
...
+    exclusive_classes: bool = False,
+    verbose: bool = False,
+):
+    """Get bulk TextCat suggestions from an OpenAI API, using zero-shot or
+    few-shot learning.  The results can then be corrected using the
+    `textcat.manual` recipe.
+
+    Here, we use ChatGPT to suggest if a particular text talks about a recipe or
+    not. You can use the parameter `--chatgpt-bias` (float, 0 to 1 inclusive) to
+    set how much we prefer seeing examples that ChatGPT thinks as recipes.
```
koaning, line 298: I think this is partially the text from the fetch recipe and it also still references `recipes`.

recipes/openai_textcat_multilabel.py
```diff
@@ -0,0 +1,527 @@
...
+                {"view_id": "html", "html_template": HTML_TEMPLATE},
+            ],
+            "show_flag": True,
+            "global_css": CSS_FILE_PATH.read_text(),
+        },
+    }
+
+
+@prodigy.recipe(
+    "textcat_multilabel.openai.fetch",
```
koaning, line 351: Do we really need two fetch recipes? The Prodigy [`textcat.correct`](https://prodi.gy/docs/recipes/#textcat-correct) recipe will adapt based on value for `--labels`. Is there a reason that cannot happen here?

recipes/openai_textcat_multilabel.py
```diff
@@ -0,0 +1,527 @@
...
+    """Get bulk TextCat suggestions from an OpenAI API, using zero-shot or few-shot learning.
+    The results can then be corrected using the `textcat.manual` recipe."""
+    api_key, api_org = _get_api_credentials(model)
+    examples = _read_prompt_examples(examples_path)
+    nlp = spacy.blank(lang)
+    if segment:
+        nlp.add_pipe("sentencizer")
+    openai = OpenAISuggester(
+        openai_model=model,
+        labels=labels,
```
koaning, line 386: When I tried running: 

```
python -m prodigy textcat_multilabel.openai.fetch doubt.jsonl doubt-annot.jsonl "sarcasm,anger,joy"  -F recipes/openai_textcat_multilabel.py
```

I got this error. 

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/prodigy/__main__.py", line 61, in <module>
    controller = recipe(*args, use_plac=True)
  File "cython_src/prodigy/core.pyx", line 329, in prodigy.core.recipe.recipe_decorator.recipe_proxy
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/plac_core.py", line 367, in call
    cmd, result = parser.consume(arglist)
  File "/home/vincent/Development/prodigy-demos/venv/lib/python3.8/site-packages/plac_core.py", line 232, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
  File "recipes/openai_textcat_multilabel.py", line 386, in textcat_multilabel_openai_fetch
    openai = OpenAISuggester(
TypeError: __init__() missing 1 required keyword-only argument: 'exclusive_classes'
```

I've manually overridden it to true on my end but I think you'll want to add this as a param for the recipe.

koaning: I also had a look at the output of this recipe and noticed something strange. 

Again, this is the recipe call: 

```
python -m prodigy textcat_multilabel.openai.fetch doubt.jsonl doubt-annot-ne.jsonl "sarcasm,anger,joy"  -F recipes/openai_textcat_multilabel.py
```

When I manually set `exclusive_classes=True` in `OpenAISuggester` then the output via `db-out` looks like this: 

```
> cat doubt-annot-excl.jsonl | jq -c "{accept: .accept}"
{"accept":["joy"]}
{"accept":["sarcasm"]}
{"accept":["joy"]}
{"accept":["sarcasm"]}
{"accept":["Sarcasm"]}
{"accept":["sarcasm"]}
{"accept":["Sarcasm"]}
{"accept":["joy"]}
{"accept":["sarcasm"]}
{"accept":["Sarcasm",""]}
...
```

When I manually set `exclusive_classes=False` in `OpenAISuggester` then the output via `db-out` looks like this: 

```
> cat doubt-annot-not-excl.jsonl | jq -c "{accept: .accept}"
{"accept":["joy","sarcasm"]}
{"accept":["sarcasm","joy",""]}
{"accept":["joy","sarcasm"]}
{"accept":["sarcasm","joy",""]}
{"accept":["sarcasm","joy"]}
{"accept":["sarcasm","joy"]}
{"accept":["sarcasm","joy"]}
{"accept":["joy","sarcasm"]}
{"accept":["sarcasm","joy",""]}
{"accept":["sarcasm","joy",""]}
...
```

It seems like it passes an empty string to the accept key here? That suggests something is going awry with parsing, but it also suggests (and Matt totally had to reminder me of this on my own recipe) that it might be good to write some unit tests that check this.

recipes/openai_textcat_multilabel.py
```diff
@@ -0,0 +1,527 @@
...
+            "global_css": CSS_FILE_PATH.read_text(),
+        },
+    }
+
+
+@prodigy.recipe(
+    "textcat_multilabel.openai.fetch",
+    input_path=("Path to jsonl data to annotate", "positional", None, Path),
+    output_path=("Path to save the output", "positional", None, Path),
+    labels=("Labels (comma delimited)", "positional", None, lambda s: s.split(",")),
```
koaning, line 354: Total nitpick, but shouldn't `labels` be an option `--labels`? Mainly to keep them consistent with the [Prodigy recipes](https://prodi.gy/docs/recipes/#textcat-correct)?

recipes/openai_textcat_multilabel.py
koaning: Oh! And also notice how the labels can have inconsistent capitalisation. Might also be something to test for.

koaning: Yeah, that those capitals are causing inconsistencies in the `textcat.manual` task that follows up.

```
python -m prodigy textcat.manual lj-review-1 doubt-annot.jsonl --label "sarcasm,anger,joy"  -F recipes/openai_textcat_multilabel.py
```

This leads to examples that are empty, like this one: 

CleanShot 2023-01-17 at 13 24 40

recipes/openai_textcat_multilabel.py
```diff
@@ -0,0 +1,527 @@
...
+    prompt_path=("Path to jinja2 prompt template", "option", "p", Path),
+    max_examples=("Max examples to include in prompt", "option", "n", int),
+    exclusive_classes=("Make the classification task exclusive", "flag", "E", bool),
+    verbose=("Print extra information to terminal", "flag", "v", bool),
+    # fmt: on
+)
+def textcat_multilabel_openai_correct(
+    dataset: str,
+    input_path: Path,
+    labels: List[str],
```
koaning, line 281: In the case of a single label, the interface looks like this:

CleanShot 2023-01-17 at 13 27 42

Wouldn't it be better to generate a binary interface for it instead? Something like: 

CleanShot 2023-01-17 at 13 28 49

Textcat manual automatically takes care of this, maybe we can do that here as well?

recipes/openai_textcat.py
```diff
@@ -0,0 +1,535 @@
...
+        for line in text.strip().split("\n"):
+            if line and ":" in line:
+                key, value = line.split(":", 1)
+                output[key.strip().lower()] = value.strip()
+        return output
+
+
+@prodigy.recipe(
+    # fmt: off
+    "textcat.openai.suggest",
```
koaning, line 272: I tried running this recipe and wondered if I would use the `fetch` command instead. 

I imagine this is somewhat a personal perference, but it seems easier to just download everything upfront first with the `fetch` recipe and to then just filter and only look at the examples where OpenAI had the class that I was interested in. This would work for both binary scenarios, as well as multi-label ones. 

@ljvmiranda921 what's your experience with this?

koaning (changes requested):
@ljvmiranda921 I really like how the reason popped up in the meta information. That was a nice touch! 

That said, I found some subtle and less-subtle bugs as I was exploring this. Some of the bugs appeared in two files, which is making me think it may be better to just put everything in a single file for now. 

Many of the bugs seem perfectly fixable, but I guess the larger discussion is about the utility of the `suggest` recipe. I wonder if the `fetch` recipe is sufficient to support for now and also less complex to support.

I may be missing some context on an earlier conversation though, so feel free to point me to any resources/discussions that can help convince me that I'm mistaken. 

My only comment on the `fetch` recipe is that it can be a bit slow (many of my datasets exceed > 10K examples) and it's not able to pick up where it left off. Any improvements here would greatly improve the experience of a user who is looking to get started right away.

recipes/openai_textcat.py
koaning: I also got this error every time I tried batch sizes of 25.

```
httpx.HTTPStatusError: Client error '400 Bad Request' for url 'https://api.openai.com/v1/completions'
```

Might be a good idea to print the response from OpenAI here. The json response usually displays the error that it's dealing with.

koaning:
Also! Do we really need to have `data/textcat_openai_fetch_dump.jsonl` in this repo?

templates/textcat_prompt.jinja2
koaning: Did you push the readme changes? I did not see them in this PR.

ljvmiranda921: Oh sorry, I meant the PR description 😅

recipes/openai_textcat_multilabel.py
ljvmiranda921: Yeah I think it might be better to just have a single fetch recipe

recipes/openai_textcat_multilabel.py
ljvmiranda921: I did a bit of refactor that combines both of them. Will update the PR description on the changes. But yes, this should be automatic now

ljvmiranda921:
Hi @koaning thanks for the review! I resolved some comments if you don't mind and here's the summary of changes. I also updated the summary in the PR description (above):

1. **Combined binary and multi textcat.** I think combining these two files and refactoring the `OpenAISuggester` solves a few problems, namely, having a single fetch recipe for both. 
2. **Weird obvious problems in parsing (e.g. empty strings, capitalization, etc.).** I added tests in `tests/test_textcat.py` with some examples. I'm not sure if I was able to replicate the empty string problem. Would you mind testing it again on your data just to be sure? I can also try it if you provide the JSONL file :) 
3. **Utility of `suggest` recipe.** At first I thought it's a good way to find false negatives, especially if you have a lot. But I realized that doing a big fetch and then randomly samping negatives afterwards should be simpler implementation-wise. I'm not too attached to `suggest` so I removed it.
4. **Custom HTML for binary classification.** Yeah I like the idea of having something like "OpenAI says: X" in the UI. I added something like that. I'd love to have it color-coded but I'm still thinking of a more elegant way to do it.
5. **On tests.** I added a few tests for parsing the response. Let me know if there are cases that I might have not seen!


### Unsolved problems

- I do encounter the slowness in `fetch`, and I'm often impeded by the retries and 50x errors. I guess retrying when I encountered them should work. I'll give that a try tomorrow or on Friday.

recipes/openai_textcat.py
ljvmiranda921: ^For the current iteration for the PR I removed the `suggest` recipe. I'm not too attached on it and I'm convinced that the `fetch` might be better (although I'm still having retry and connection problems with it at longer batches). 

Given that (assuming we only have `textcat.openai.correct` and `fetch`), then I think modeling this problem should be something like:

```sh
python3 -m prodigy textcat.openai.correct demo-doubt \
    doubt.jsonl \
   templates/textcat_prompt.jinja2 \
   --labels excitement,anger  \
   --exclusive-classes
```

recipes/openai_textcat.py
ljvmiranda921: Still working on this, but just noting that I saw this!

ljvmiranda921:
Quick update on the call errors, I added some retry and exponential backoff so that the user experience is better (before it looks seemingly stuck, but after printing the timeout at least the user knows that something's happening)

image

recipes/openai_textcat.py
ljvmiranda921: Updates: https://github.com/explosion/prodigy-openai-recipes/pull/50#issuecomment-1396563787

ljvmiranda921:
PR for NER demo: https://github.com/ljvmiranda921/prodigy-openai-recipes/pull/1

koaning:
We talked about this during our huddle, but I figured sharing here as well. The OpenAI devrel mentioned an update to the [rate limit guide](https://beta.openai.com/docs/guides/rate-limits/what-are-the-rate-limits-for-our-api). It seems new users on their platform only get to have 60 requests a second in the first 48 hours of the account. 

We should definitely document this, but it could also be a good idea to mention this as a warning in the terminal. Just so folks know it's not the recipe that's the issue.

recipes/openai_textcat.py
```diff
@@ -0,0 +1,287 @@
...
+
+    # Create partial render of the template
+    if not exclusive_classes and len(labels) == 1:
+        msg.warn(
+            "Binary classification should always be exclusive. Setting "
+            "`exclusive_classes` parameter to True"
+        )
+        exclusive_classes = True
+
+    template = Template(
```
koaning, line 166: Doesn't `load_template` always return a `Template` object already?

recipes/openai_textcat.py
```diff
@@ -0,0 +1,287 @@
...
+        openai_model=model,
+        openai_timeout_s=30,
+        openai_n=10,
+        verbose=verbose,
+    )
+    for eg in examples:
+        openai.add_example(eg)
+
+    # Set up the stream
+    stream = list(srsly.read_jsonl(input_path))
```
koaning, line 189: Technically, the user might provide a CSV as well here, right?

The [`get_stream`](https://prodi.gy/docs/api-components#get_stream) function would be better here.

recipes/openai.py
```diff
@@ -0,0 +1,354 @@
...
+    with an error code in `error_codes`. The amount of timeout also increases
+    exponentially every time we retry.
+    """
+    assert n >= 0
+    assert timeout_s >= 1
+    r = call_api()
+    i = -1
+    # We don't want to retry on every non-ok status code. Some are about
+    # incorrect inputs, etc. and we want to terminate on those.
+    while i < n and r.status_code in error_codes:
```
koaning, line 324: I noticed httpx.ReadTimeout errors as well. They seemed to break the flow just around 60 requests. 

```
Traceback (most recent call last):
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_exceptions.py", line 10, in map_exceptions
    yield
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/backends/sync.py", line 28, in read
    return self._sock.recv(max_bytes)
  File "/usr/lib/python3.8/ssl.py", line 1226, in recv
    return self.read(buflen)
  File "/usr/lib/python3.8/ssl.py", line 1101, in read
    return self._sslobj.read(len)
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_transports/default.py", line 60, in map_httpcore_exceptions
    yield
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_transports/default.py", line 218, in handle_request
    resp = self._pool.handle_request(req)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/connection_pool.py", line 253, in handle_request
    raise exc
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/connection_pool.py", line 237, in handle_request
    response = connection.handle_request(request)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/connection.py", line 90, in handle_request
    return self._connection.handle_request(request)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/http11.py", line 112, in handle_request
    raise exc
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/http11.py", line 91, in handle_request
    ) = self._receive_response_headers(**kwargs)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/http11.py", line 155, in _receive_response_headers
    event = self._receive_event(timeout=timeout)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_sync/http11.py", line 191, in _receive_event
    data = self._network_stream.read(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/backends/sync.py", line 28, in read
    return self._sock.recv(max_bytes)
  File "/usr/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc)
httpcore.ReadTimeout: The read operation timed out

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/prodigy/__main__.py", line 61, in <module>
    controller = recipe(*args, use_plac=True)
  File "cython_src/prodigy/core.pyx", line 364, in prodigy.core.recipe.recipe_decorator.recipe_proxy
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/plac_core.py", line 367, in call
    cmd, result = parser.consume(arglist)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/plac_core.py", line 232, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
  File "recipes/openai_textcat.py", line 287, in textcat_openai_fetch
    srsly.write_jsonl(output_path, stream)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/srsly/_json_api.py", line 138, in write_jsonl
    for line in lines:
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/recipes/openai.py", line 129, in set_hashes
    for example in stream:
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/recipes/openai.py", line 175, in format_suggestions
    for example in stream:
  File "cython_src/prodigy/components/preprocess.pyx", line 164, in add_tokens
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/spacy/language.py", line 1545, in pipe
    for doc in docs:
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/spacy/language.py", line 1589, in pipe
    for doc in docs:
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/spacy/language.py", line 1586, in <genexpr>
    docs = (self._ensure_doc(text) for text in texts)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/spacy/language.py", line 1535, in <genexpr>
    docs_with_contexts = (
  File "cython_src/prodigy/components/preprocess.pyx", line 157, in genexpr
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/recipes/openai.py", line 159, in stream_suggestions
    responses = self._get_openai_response(prompts)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/recipes/openai.py", line 196, in _get_openai_response
    r = retry(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/recipes/openai.py", line 320, in retry
    r = call_api()
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/recipes/openai.py", line 197, in <lambda>
    lambda: httpx.post(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_api.py", line 304, in post
    return request(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_api.py", line 100, in request
    return client.request(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_client.py", line 821, in request
    return self.send(request, auth=auth, follow_redirects=follow_redirects)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_client.py", line 908, in send
    response = self._send_handling_auth(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_client.py", line 936, in _send_handling_auth
    response = self._send_handling_redirects(
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_client.py", line 973, in _send_handling_redirects
    response = self._send_single_request(request)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_client.py", line 1009, in _send_single_request
    response = transport.handle_request(request)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_transports/default.py", line 218, in handle_request
    resp = self._pool.handle_request(req)
  File "/usr/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/vincent/Development/prodigy-demos/openai-prodigy-recipes/venv/lib/python3.8/site-packages/httpx/_transports/default.py", line 77, in map_httpcore_exceptions
    raise mapped_exc(message) from exc
httpx.ReadTimeout: The read operation timed out
```

templates/textcat_prompt.jinja2
```diff
@@ -0,0 +1,38 @@
+{% if labels|length == 1 %}
+    {% set label = labels[0] %}
```
koaning, line 2: Is the indentation here needed? It seems to re-appear in the prompt. 

```
╭──────────────────────────────────────────────────────── Prompt to OpenAI ────----------------------------------------───────────────╮
│                                                                                                                                     │
│     From the text below, classify if it belongs to the following labels: website purchase, complaint, delivery related, compliment  │
│                                                                                                                                     │
│     The task is non-exclusive, so you can provide more than one label as long as                                                    │
│     they're comma-delimited. For example: Label1, Label2, Label3.                                                                   │
│                                                                                                                                     │
│                                                                                                                                     │
│                                                                                                                                     │
│ Answer in the following format:                                                                                                     │
│                                                                                                                                     │
│ answer: <string>                                                                                                                    │
│ reason: <string>                                                                                                                    │
│                                                                                                                                     │
│ Text:                                                                                                                               │
│ """                                                                                                                                 │
│ @tesco having to order from Sainsbury's as your site doesn't allow me to change my current home address. https://t.co/0ZsikSrVcd    │
│ """                                                                                                                                 │
│                                                                                                                                     │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

File: recipes/openai_textcat.py [hunk 19]
```diff
@@ -0,0 +1,287 @@
...
+        output = self._parse_output(response)
+        example["answer"] = output["answer"].lower() == "accept"
+        example["meta"] = {
+            "answer": output["answer"].upper(),
+            "reason": output["reason"],
+        }
+        example["label"] = self.labels[0]
+        return example
+
+    def _parse_multi(self, example: Dict, response: str) -> Dict:
```
koaning, line 104: I have a file called `one-off.jsonl` with this example in it:

```
{"text":"@Tesco just for completeness ordered arrived just now... over an hour late https:\/\/t.co\/NGydnyAm6Q"}
```

Next, I tell OpenAI to annotate it. 

```
python -m prodigy textcat.openai.fetch one-off.jsonl templates/textcat_prompt.jinja2 one-off-annot-v2.jsonl --label "website purchase,complaint,delivery related,compliment" --batch-size 10 --examples-path examples.yaml --verbose -F recipes/openai_textcat.py
```

Then, when I look in the annotated json, these are the contents:

```
{
  "text": "@Tesco just for completeness ordered arrived just now... over an hour late https://t.co/NGydnyAm6Q",
  "openai": {
    "prompt": "\n    From the text below, classify if it belongs to the following labels: website purchase, complaint, delivery related, compliment\n    \n    The task is non-exclusive, so you can provide more than one label as long as\n    they're comma-delimited. For example: Label1, Label2, Label3.\n    \n\n\nAnswer in the following format:\n\nanswer: <string>\nreason: <string>\n\nText:\n\"\"\"\n@Tesco just for completeness ordered arrived just now... over an hour late https://t.co/NGydnyAm6Q\n\"\"\"\n",
    "response": "\nAnswer: Complaint, Delivery related\nReason: The text contains a mention of Tesco, which is a website purchase, as well as a mention of the delivery being late, which is a complaint and delivery related."
  },
  "tokens": [...],
  "meta": {
    "reason": ""
  },
  "options": [
    {
      "id": "website purchase",
      "text": "website purchase"
    },
    {
      "id": "complaint",
      "text": "complaint"
    },
    {
      "id": "delivery related",
      "text": "delivery related"
    },
    {
      "id": "compliment",
      "text": "compliment"
    }
  ],
  "answer": "accept",
  "accept": [],
  "_input_hash": 1284924640,
  "_task_hash": 1095851038
}
```

The answer is empty, and so is the reason. I've seen this happen to a bunch of my examples, so I think something may be up with the parsing.

koaning: Yeah, I just checked the tests and it seems like the `test_parser_response_*` tests are failing. 

That suggests to me that we also want to make sure that the CI actually runs on this PR now. @ljvmiranda921 I just added you as a writer to this repo, could you make a new PR that isn't forked from your repo? That should allow Prodigy to download in the CI and the tests to run.

recipes/openai_textcat.py
```diff
@@ -0,0 +1,287 @@
...
+    # Create OpenAISuggester with ChatGPT parameters
+    openai = TextCatOpenAISuggester(
+        prompt_template=template,
+        labels=labels,
+        max_examples=max_examples,
+        segment=segment,
+        openai_api_org=api_org,
+        openai_api_key=api_key,
+        openai_model=model,
+        openai_timeout_s=10,
```
koaning, line 277: I wonder if there's any harm in increasing this to 20s. Is there a reason the timeout is 10s?

koaning (changes requested):
I fear the tests aren't passing and the parsing is currently broken. I might do a deep dive later today for some details, but for the more pressing thing seems to actually have the CI run so we can catch this before a review starts.

ljvmiranda921:
Ack thanks for the comment. Wasn't able to push the fixing commit. Yeah you're right it's better if I PR not from my fork so that the CI shows up. Will close this PR now and open up another

recipes/openai_textcat.py
ljvmiranda921: The problem with this unfortunately is that I want to partially render the template 🤔 I don't want to pass `exclusive_classes` nor `labels` to the main suggester class. I'll think of a more elegant ways to do this (maybe via kwargs or something)

recipes/openai_textcat.py
ljvmiranda921: Noted on this!

PR #51, opened 2023-01-13
OpenAI terms recipes

koaning:
I've started work on an OpenAI recipe to generate term phrases. It's a work in progress but it seems to work very well! Here's a few examples you could run yourself. 

```bash
# Base behavior, fetch at least 100 terms/phrases
python -m prodigy terms.openai.fetch "boardgames" boardgames.jsonl --n 100 -F recipes/openai_terms.py

# Fetch 250 phrases, but give a sign of progress
python -m prodigy terms.openai.fetch "xbox games" xbox.jsonl --n 250 --progress -F recipes/openai_terms.py

# Supply some seeds yourself
python -m prodigy terms.openai.fetch "scateboard tricks" scateboard.jsonl --n 250 --seeds "kickflip,ollie" -F recipes/openai_terms.py

# Add 50 more examples and resume from the `.jsonl` file. This is slow! 
python -m prodigy terms.openai.fetch "scateboard tricks" scateboard.jsonl --n 50 --resume -F recipes/openai_terms.py
```

The prompt that I am using for this is relatively simple. 

```
 Generate me 50 examples of shoe brands.                                                                                           
                                                                                                                                   
 Here are the examples:                                                                                                            
 - adidas                                                                                                                          
 - nike                                                                                                                            
 - converse                                                                                                                        
 - vans                                                                                                                                                                                                                                                  
 -                                                                                                                                                              ```

### CODE REVIEW

**Comment by koaning** [comment]
We can add some seeds to the list to help steer OpenAI but it's basically just a list that we get back. The recipe extends the list on every request and after each request I append the new phrases to the output file.  

To keep it consistent with the Prodigy terms recipe, I figured I'd add a `--resume` flag, but this is terribly slow. OpenAI will try to make sure that every item added to the list is unique, but if the current example list is very large ... it will only make one suggestions per request. That's why I'm wondering if it's best to remove the `--resume` flag for now. 

I will say, as-is, this seems pretty useful. And if folks agree I will proceed by adding the `terms.openai.correct` and `terms.openai.to-patterns` recipes. I'm also wondering if this recipe should be called `terms.openai.fetch`, maybe `phrases.openai.fetch` is more appropriate?

File: recipes/openai_terms.py [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+
+
+def _parse_terms(completion):
+    # Sometimes it only returns a single item. For example, when there are
+    # many, many seeds around.
+    if "\n" not in completion:
+        return [completion.replace("-", "").strip().lower()]
+    # Other times we cannot assume the final item will have had sufficient 
+    # tokens available to complete the term, so we have to discard it.
+    lines = completion.split("\n")
```
honnibal, line 48: Do we need a `strip()` here? Or are we sure there can't be trailing newlines?

recipes/openai_terms.py
```diff
@@ -0,0 +1,122 @@
...
+            "https://api.openai.com/v1/completions",
+            headers=headers,
+            json={
+                "model": model,
+                "prompt": [prompt],
+                "temperature": 1,
+                "max_tokens": max_tokens,
+            },
+            timeout=30,
+        )
```
honnibal, line 109: This should be batched, it's too slow at one item per query.

recipes/openai_terms.py
koaning: I suppose `.lstrip()` should suffice.

koaning:
TODO: 

- [ ] refactor `_parse_terms`, it currently doesn't pass DRY 
- [ ] add `temperature` parameter and figure out sensible starting param

recipes/openai_terms.py
```diff
@@ -0,0 +1,122 @@
...
+    headers = {
+        "Authorization": f"Bearer {os.getenv('OPENAI_KEY')}",
+        "OpenAI-Organization": os.getenv("OPENAI_ORG"),
+        "Content-Type": "application/json",
+    }
+    while len(terms) < n:
+        prompt = template.render(n=n, examples=seeds + terms, description=query)
+        if verbose:
+            rich.print(Panel(prompt, title="Prompt to OpenAI"))
+        resp = httpx.post(
```
ljvmiranda921, line 99: In case you're encountering problems with timeouts (not the 429 error), you might try using a [`httpx.Client`](https://www.python-httpx.org/advanced/) to make the `POST` call.

recipes/openai_terms.py
```diff
@@ -0,0 +1,122 @@
...
+    terms = []
+    if resume:
+        examples = srsly.read_jsonl(output_path)
+        seeds.extend([e["text"] for e in examples])
+    headers = {
+        "Authorization": f"Bearer {os.getenv('OPENAI_KEY')}",
+        "OpenAI-Organization": os.getenv("OPENAI_ORG"),
+        "Content-Type": "application/json",
+    }
+    while len(terms) < n:
```
ljvmiranda921, line 95: I think the general approach looks good! 

I don't think it's necessary for now but we can also take a look at reusing the `OpenAI*` class in the `openai_ner` template. I understand that it might complicate stuff right now but we're thinking about refactoring the structure and having all recipes call that adaptor class.

koaning: I had a look at that class but had the impression that it wouldn't work for this recipe. For example, the `labels` parameter doesn't really make sense here. 

I'm definitely in favor of consolidating some of the building blocks and that we reuse components. But I think the class as-is is more designed for active-y learning-y tasks in Prodigy while this recipe describes much more of a one-off thing you'd run.

recipes/openai_terms.py
koaning: I just tried with just "lstrip" and found whitespaces at the end. Will stick to `.strip()`.

koaning:
The `--resume` setting is no longer slow now that I've added batching. I was intially worried about duplicates, but it seems the extra batches more than compensate for that. Yay!

recipes/openai_terms.py
```diff
@@ -0,0 +1,159 @@
...
+                    "n": min(n_batch, best_of),
+                    "best_of": best_of
+                },
+                timeout=30,
+            )
+            choices = resp.json()['choices']
+        except KeyError:
+            msg.fail("Did not receive completions from OpenAI. Instead we received:")
+            rich.print(resp.json())
+            exit(code=1)
```
honnibal, line 137: This is catching `choices` not being available in the response json, right?

An `if` statement over the response code would be better here.

koaning: Fair call. Made the change!

koaning:
I noticed that the `--seed` terms weren't added to the output file. The original `terms.teach` recipe does do this, so I adapted the recipe to immediately add the seed terms to the output file if they aren't in there yet.

koaning:
Magda had a good feature request during her review: we should also generate variants of the terms that we generate.

So I added this feature but in a new recipe: `terms.openai.variants`. This recipe will take parents from a previously fetched terms file and allow you to generate variants for them. The reason for making a new recipe is that this way we don’t have to worry about the OpenAI token limits. We can also show the parent in the meta information of the example and it also allows us to --resume where we left off. The prompt looks like this:

```
Generate the alternative spellings for {{example}}. It's a {{description}}. Make sure everything that is generated is a {{description}}.

Examples:
- {{example}}
-
```

I found that when you add a description twice it doesn’t just go and abbreviate everything. I also learned that the `frequency_penalty` hyperparam matters for this one.

## Try it? 

If you’d like to give this recipe a spin, you’ll first want to download a term list to start with.

```
python -m prodigy terms.openai.fetch "xbox games" terms-xbox-start.jsonl --progress -F recipes/openai_terms.py
```

And from here, you can generate variations.

```
python -m prodigy terms.openai.variants "xbox games" terms-xbox-start.jsonl terms-xbox-extra.jsonl --progress --resume  -F recipes/openai_terms.py
```

If people like this approach I can start writing the docs for it.

README.md
```diff
@@ -159,6 +159,120 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+| `--max-tokens`, `-mt` | int   | Max tokens to generate per call                     | `100`                           |
+
+### Example usage
+
+Suppose you're interested in detecting skateboard tricks in text, then you might want to start
+with a term list of known tricks. You might want to start with the following query:
+
+```bash
+# Base behavior, fetch at least 100 terms/phrases
+python -m prodigy terms.openai.fetch "skateboard tricks" tricks.jsonl --n 100 -F recipes/openai_terms.py
```
magdaaniol, line 195: You should probably specify the template on the command line in this example. If run as is, it uses the variation template that is hardcoded as default in the current version of the recipe.

recipes/openai_terms.py
```diff
@@ -0,0 +1,379 @@
...
+
+        # Cast to a set to make sure we remove duplicates
+        choices = resp.json()["choices"]
+        sets_of_terms = [set(_parse_terms(c["text"])) for c in choices]
+        parsed_terms = list(reduce(lambda a, b: a.union(b), sets_of_terms))
+
+        # Save intermediate results into file, in-case of a hiccup
+        srsly.write_jsonl(
```
magdaaniol, line 228: I would either write to  the final file outside the loop trimming the output to `n` or document that the recipe will likely overshoot the `n` target.

koaning:
@honnibal I've taken Magda's comments into account and I removed the variants stuff. This code is in a private repo on my end, so I can always have another look at it. But this keeps the terms recipe simple and it's already plenty useful. 

Any concerns? If not, can I merge 😄 ?

koaning:
There was a merge conflict, took care of it. Will merge once CI is green.

PR #52, opened 2023-01-14
Minor cleanup of NER recipe. 

koaning:
`load_dotenv()` was called twice. This PR cleans that.


ljvmiranda921 (approved):
I think we can merge this now :) Unless there are other cleanup stuff we'd like to add?

koaning:
There may be other cleanups possible, but this was such a low-hanging one I figured just making the PR immediately.

PR #53, opened 2023-01-19
Paraphrasing Recipe

koaning:
This PR adds support for paraphrasing. It's similar to the terms recipe, but it's slightly different. It's meant to generate variants of examples that exist in a existing file. There are many ways to "order a pizza" and it can be a good idea to see if OpenAI can explore some of the edge cases.


koaning:
It seems to work. Let's say you've got a file with examples:

```json
{"text": "Can I please have a cup of cofee with a happy meal?"}
{"text": "I want a double-cheeseburger with fries and a large soda"}
{"text": "Four burgers, four fries and six mcnuggets"}
{"text": "Just a Coke please."}
{"text": "Two breakfast oatmeals."}
{"text": "Seven burgers, two fries, 6 sodas, a side order of union rings and six mcnuggets"}
```

It's important for this file to contain examples that are somewhat diverse, otherwise you might as well use the `terms` recipe. From here, the paraphrase recipe can adapt these examples. 

```bash
# Example call
python -m prodigy paraphrase.openai.fetch "MacDonalds breakfast orders" examples.jsonl paraphrased-macdo.jsonl --progress -ne 5 -F recipes/openai_paraphrase.py
```

Here's an example prompt sent to OpenAI.

```
╭──────────────────────────--------- Prompt to OpenAI ───-------------───-------------------╮
│ I want to generate more examples of MacDonalds breakfast orders. Create some variants of  |
| the following sentences.                                                                  |
│                                                                                           │
│ Examples:                                                                                 │
│ - Can I please have a cup of cofee with a happy meal?                                     │
│ - Two breakfast oatmeals.                                                                 │
│ - I want a double-cheeseburger with fries and a large soda                                │
│ - Four burgers, four fries and six mcnuggets                                              │
│ - Sever burgers, two fries, 6 sodas, a side order of union rings and six mcnuggets        │
│                                                                                           │
│ Variants:                                                                                 │
│ -                                                                                         │
╰───────────────────────────────────────────────────────────────────────────────────────────╯
```

Here are some of the examples that it generates:

```json
{"text":"Two big macs, three small sodas and two orders of onion rings","meta":{"openai_query":"MacDonalds breakfast orders"}}
{"text":"A McMuffin sandwich with a sausage patty, hash browns and a medium iced latte","meta":{"openai_query":"MacDonalds breakfast orders"}}
{"text":"Five Big Macs, two orders of small fries and eight mcnuggets","meta":{"openai_query":"MacDonalds breakfast orders"}}
{"text":"Three Big Macs, four apple pies and six mcnuggets","meta":{"openai_query":"MacDonalds breakfast orders"}}
{"text":"A medium order of french fries and two sixpiece chicken nugget meals","meta":{"openai_query":"MacDonalds breakfast orders"}}
{"text":"I would like a Big Mac Meal with large fry and a large Dr. Pepper","meta":{"openai_query":"MacDonalds breakfast orders"}}
```

ljvmiranda921:
I'm wondering if there's a way to make this functionality leverage Prodigy more. From its current usage, this can exist even without interfacing with Prodigy. Perhaps we can also do something like:

- Take an already annotated Prodigy dataset and generate `n` synthetic samples from a portion of it.
- Automatically include the synthetic samples back to the stream. So I can just pass 100 gold-annotated data (because maybe that's what the user only has), then from there let Prodigy generate `n` more synthetic examples. Then we communicate through the UI which one's from the original dataset and which one's generated by ChatGPT (the synthetic examples may even be pre-annotated). We can even extend this to any task like `ner`, `textcat` etc.

Other UI thoughts:
- I like the "flagging" mechanism in NER. We can also implement it here. More like: "hey, I like this example you generated, can you do something similar with this type?"

koaning:
> Straight up synthetic data generation.

I was thinking about keeping a text classification label intact from the parent example and attaching that to a synthetic example but unfortunately that introduces a bunch of problems. 

1. OpenAI can really get it wrong sometimes. So assuming that the label is still appropriate feels dangerous.
2. You might want to paraphrase a sentence and be interested in entities too. If you want to attach these labels automatically it'd be better to pass it to the `ner.openai.fetch`/`textcat.openai.fetch` recipe and manually label after that. 

That said, you do make a fair point about targeting a Prodigy dataset instead of a `.jsonl` file on disk. The recipe should totally be able to accept a dataset from the command line. Will add that to a TODO! 

> Automatically include the synthetic samples back to the stream.

I already fear that the synthetic examples that are generate can be "too synthetic". By synthesizing data that based on already synthesized data I worry that we can go too far into the synthetic route. 

There's a balancing act between generating more data vs. having a dataset that is representative of what actual users might generate. It can be a good thing to generate a larger training set, but if the synthetic data is unlike the user data, then it might lead to bias in the model.

> Flag good examples?

In my mind this generated dataset would be input for Prodigy, so good examples would be accepted like they would normally with any other training data. These "good" examples could then become part of the standard dataset, which theoretically could then be used by the recipe again. The only downside here is that you'd have the "maybe too synthetic"-problem again.

koaning:
TODO: 

- [] Allow this recipe to also read in datasets from Prodigy. We should ensure that only accepted examples get taken into account.

koaning:
There are a few places where I felt urged to raise a warning. 

- In [`prodigy/components/feeds.py`](https://github.com/explosion/prodigy/blob/main/prodigy/components/feeds.py) we have [the main culprit](https://github.com/explosion/prodigy/blob/main/prodigy/components/feeds.py#L328) which this PR raises a warning for. 

- In [`prodigy/recipes/data_utils.py`](https://github.com/explosion/prodigy/blob/main/prodigy/recipes/data_utils.py) we have a helper
called `load_examples` that is used inside of `get_train_eval_examples`. Inside `load_examples` we rehash and overwrite any example that isn't ignored. This shouldn't be needed because we're loading in a dataset from Prodigy, which should already have hashes. So I figured adding a warning couldn't hurt.

But there are a few places where it felt fine. 

- In [`prodigy/components/loaders.py`](https://github.com/explosion/prodigy/blob/main/prodigy/components/loaders.py) we use `set_hashes` in `get_stream`. There are a few edge cases where this might cause issues (see [slack discussion](https://explosion-ai.slack.com/archives/C0241LQED4G/p1674216888452409)), but this feels outside of the scope of this PR. 

- In [`prodigy/components/preprocess.py`](https://github.com/explosion/prodigy/blob/main/prodigy/components/preprocess.py) we use `set_hashes` in `split_sentences`, `split_spans`, `add_labels_to_stream`. These all seem reasonable to me because we're creating new examples and we want to make sure that these new examples have appropriate hashes.

- In [`matcher.py`](https://github.com/explosion/prodigy/blob/main/prodigy/models/matcher.py) we reset the hashes when a "prediction" is made in the `__call__` method of the `PatternMatcher`. This is because recipes that use the pattern-matcher might treat an example with/without spans as different. That feels fine.

- In [`ner.py`](https://github.com/explosion/prodigy/blob/main/prodigy/models/ner.py) we set hashes in the `merge_spans` helper, which feels fine as, like before, we're creating new examples and we want to make sure that these new examples have appropriate hashes. We also set hashes when spans are added for the best predictions in `EntityRecognizer.make_best`. It also appears in `get_tasks` which is defined in the `predict` function. It feels fine here, because we set the hashes as a result of making a prediction. 

- In [`tagger.py`](https://github.com/explosion/prodigy/blob/main/prodigy/models/tagger.py) we see something very similar to the `ner.py` file. The `merge_tags` helper uses it and `Tagger.__call__` uses it for similar purposes. 

- We use `set_hashes` in a bunch of recipes and I'd say these are all mostly fine because they're used explicitly, which is what's supposed to happen. I do worry that some recipes like `db-in`/`db-merge` might rehash examples that have a task/input key missing while assigning a task/input hash. But this is related to the aforementioned [slack discussion](https://explosion-ai.slack.com/archives/C0241LQED4G/p1674216888452409) and outside of the scope of this PR.

PR #54, opened 2023-01-19
Add TextCat recipes

ljvmiranda921:
## Description

This Pull Request does two things:

- Add two recipes for text classification with the help of ChatGPT
- Refactor the `OpenAISuggester` as a base class so that it's more reusable

## Add recipes for text classification

This PR adds two recipes, `textcat.openai.correct` and `textcat.openai.fetch`, to perform text classification in a streaming and batch manner. I also removed the  `textcat.openai.suggest` because the `fetch` command already fulfills the same purpose.

Unlike the previous state of this PR, I have now combined binary and multiclass textcat into a single recipe, where the logic is based on the number of labels (i.e., len(labels)>1 then multiclass). The template also follows a similar logic using if-else statements found in Jinja2. 

I've also made the `prompt_path` a positional argument because I think it reads better, i.e.:

> Help me annotate <input_path> using this <prompt_path>

Below are a few examples:

### Example 1: Binary classification

In this example, we perform binary classification by passing a single label to `--labels`. The `view_id` also changes with respect to this value.

```sh
python3 -m prodigy textcat.openai.correct textcat_test \
    data/reddit_r_cooking_sample.jsonl \ # input path
    --prompt-path templates/textcat_prompt.jinja2 \ # template path
    --labels recipe
    -F recipes/openai_textcat.py
    --verbose
```

### Example 2: Multiclass / Multilabel classification

For multiclass (exclusive) classification, we pass the parameter
`--exclusive-classes` or `-E` to state that we only want a single value. The
`config.choice` parameter changes accordingly:

```sh
python3 -m prodigy textcat.openai.correct textcat_test \
    data/reddit_r_cooking_sample.jsonl \ # input path
    --prompt-path templates/textcat_prompt.jinja2 \ # template path
    --labels recipe,feedback,question
    -F recipes/openai_textcat.py
    --exclusive-classes
    --verbose
```

For multilabel classification, we don't set the `-E` flag. Notice that it's now a checkbox instead of a choice button:

```sh
python3 -m prodigy textcat.openai.correct textcat_test \
    data/reddit_r_cooking_sample.jsonl \ # input path
    --prompt-path templates/textcat_prompt.jinja2 \ # template path
    --labels recipe,feedback,question
    -F recipes/openai_textcat.py
    --exclusive-classes
    --verbose
```

## Refactor OpenAISuggester

I went ahead and attempted to generalize the `OpenAISuggester` class. It is now located in `recipes/openai.py`. It is now an abstract base class (`abc.ABC`) and has an abstractmethod `parse_response`. This means that we need to inherit `OpenAISuggester` and implement the `parse_response(example: Dict, response: str)` function.

The role of `parse_response` is to convert whatever ChatGPT returns (`response`) and assign
the parsed values into appropriate keys in a copy of `example`. You can see an example implementation of this in `recipes:openai_textcat.py:TextCatOpenAISuggester`. 

## Changelog
- January 18: did a few refactors after initial review.


recipes/openai_textcat.py
```diff
@@ -0,0 +1,287 @@
...
+    if not exclusive_classes and len(labels) == 1:
+        msg.warn(
+            "Binary classification should always be exclusive. Setting "
+            "`exclusive_classes` parameter to True"
+        )
+        exclusive_classes = True
+
+    template = Template(
+        load_template(prompt_path).render(
+            exclusive_classes=exclusive_classes, labels=labels, examples=examples
```
ljvmiranda921, line 168: The problem with this unfortunately is that I want to partially render the template 🤔 Partial rendering doesn't work for logical parts of jinja (for loops, etc.). I don't want to pass `exclusive_classes` nor `labels` to the main suggester class. I'll think of a more elegant ways to do this (maybe via kwargs or something)

recipes/openai_textcat.py
```diff
@@ -0,0 +1,287 @@
...
+        openai_model=model,
+        openai_timeout_s=30,
+        openai_n=10,
+        verbose=verbose,
+    )
+    for eg in examples:
+        openai.add_example(eg)
+
+    # Set up the stream
+    stream = list(srsly.read_jsonl(input_path))
```
ljvmiranda921, line 189: Noted on using the `get_stream` function!

recipes/openai_textcat.py
koaning: Ahhh wait. Now I see. 

If it's a partial rendering it makes complete sense, but maybe add a comment that makes it clear to the reader.

recipes/openai.py
```diff
@@ -0,0 +1,354 @@
...
+    with an error code in `error_codes`. The amount of timeout also increases
+    exponentially every time we retry.
+    """
+    assert n >= 0
+    assert timeout_s >= 1
+    r = call_api()
+    i = -1
+    # We don't want to retry on every non-ok status code. Some are about
+    # incorrect inputs, etc. and we want to terminate on those.
+    while i < n and r.status_code in error_codes:
```
ljvmiranda921, line 324: > I noticed httpx.ReadTimeout errors as well. They seemed to break the flow just around 60 requests.

My gut fix would be to increase the timeout. But yeah, I'll figure out why it happens with just 60 requests.

templates/textcat_prompt.jinja2
```diff
@@ -0,0 +1,38 @@
+{% if labels|length == 1 %}
+    {% set label = labels[0] %}
```
ljvmiranda921, line 2: > Is the indentation here needed? It seems to re-appear in the prompt.

Ah agree it's not needed and didn't intend it there. Just got used to the indentation within if else. Will fix this tomorrow

recipes/openai.py
```diff
@@ -0,0 +1,354 @@
...
+    ) -> Iterable[Dict]:
+        if self.segment:
+            stream = preprocess.split_sentences(nlp, stream)
+
+        stream = self.pipe(stream, nlp, batch_size, **kwargs)
+        stream = self.set_hashes(stream)
+        return stream
+
+    @abc.abstractmethod
+    def parse_response(self, example: Dict, response: str, nlp: Language) -> Dict:
```
ljvmiranda921, line 111: > Right now `parse_response` is implemented as an abstract base class that needs to be implemented. That could work. > But wouldn't it be less effort for users if instead the function that parses the response is passed as input? That way a 
> user might implement something like:

```python
dutch_textcat_suggester = OpenAISuggester(
    prompt_template=dutch_prompt_template,
    response_parser=dutch_response_parser,
    self.labels=["positief", "neutraal", "negatief"]
)
```

I think that's a good design! Sure will do that 👍

recipes/openai_textcat.py
```diff
@@ -0,0 +1,287 @@
...
+    # Create OpenAISuggester with ChatGPT parameters
+    openai = TextCatOpenAISuggester(
+        prompt_template=template,
+        labels=labels,
+        max_examples=max_examples,
+        segment=segment,
+        openai_api_org=api_org,
+        openai_api_key=api_key,
+        openai_model=model,
+        openai_timeout_s=10,
```
ljvmiranda921, line 277: > I wonder if there's any harm in increasing this to 20s. Is there a reason the timeout is 10s?

No particular reason! Yeah I agree that making it `20s` should work. 👍

ljvmiranda921 (commented):
I copied over some review comments and my replies from the previous PR :)

recipes/openai.py
koaning: I'm finding these errors to be pretty random. They can appear all of the sudden and go away quickly.

koaning:
Mhm. This is akward. 

I called OpenAI with: 

```
python -m prodigy textcat.openai.fetch cluestarred.jsonl templates/textcat_prompt.jinja2 tesco-annot-v3.jsonl --label "website purchase,complaint,delivery related,compliment" --batch-size 10 --examples-path examples.yaml --n-examples 3 --verbose -F recipes/openai_textcat.py
```

In this case, this is the `examples.yaml` file:

```
- text: "The delivery was great. This has been great. Thanks! https://twitter.com"
  answer: "compliment,delivery related"
  reason: "This is a very general compliment about a delivery. It's not about the website, even though a website appears."
- text: "Good grief. I tried ordering something from the website and it failed."
  answer: "website purchase,complaint"
  reason: "This is complaint about an order that faired on the website."
- text: "Tesco are complaints taken seriously? Do you take customer complaints seriously or just pay it lip service???"
  answer: "complaint"
  reason: "This is complaint but it's about customer service, not about deliveries."
```

And yet, it gets it wrong on an examples that's in the examples file. 

CleanShot 2023-01-19 at 16 21 13

I don't know if there's really much we can do here but accept that OpenAI can really get it wrong. I'd like to understand if there's something we might do to steer. Documenting that would sure help users.

ljvmiranda921:
I observe that adding examples makes the prompt output harder to steer, and in some cases, results to weird outputs. I tried your example and a combination of "always answer in the following format:" etc. etc. and it does some weird things like generating new text. I agree that we should document it.

recipes/openai_textcat.py
```diff
@@ -0,0 +1,283 @@
...
+        full_text = example["text"]
+        reason = example["meta"].get("reason")
+        if len(labels) == 1:
+            answer = "accept" if example.get("answer") else "reject"
+        else:
+            answer = example.get("accept")
+        return cls(text=full_text, answer=answer, reason=reason)
+
+
+def make_textcat_response_parser(labels: List[str]) -> Callable:
```
ljvmiranda921, line 67: The parser implementation is now a function instead of inheriting some base class. So we have a `make_textcat_response_parser` return a function that performs the parsing. 

I'd want to do the same pattern with `PromptExample`, but still figuring out the best way to do so.

recipes/openai.py
```diff
@@ -0,0 +1,378 @@
...
+        openai_api_org: str,
+        openai_api_key: str,
+        openai_model: str,
+        response_parser: Callable,
+        openai_temperature: int = 0,
+        openai_max_tokens: int = 500,
+        openai_retry_timeout_s: int = 1,
+        openai_read_timeout_s: int = 30,
+        openai_n_retries: int = 10,
+        openai_n: int = 1,
```
ljvmiranda921, line 109: I also made a few parameters more explicit. For example:
- `openai_retry_timeout_s`: is the amount of time to wait before trying the call again. This scales exponentially, i.e., 2x for every retry.
- `openai_read_timeout_s`: basically the amount of time to wait before passing a `httpx.ReadTimeOut` error. I increased this to 30 seconds.
- `openai_n_retries`: to avoid confusion, this parameter replaces `openai_n` (which dictates how many results to generate). This one controls how many times we will retry the call in the case of 429 errrors.
- `openai_n`: this one is more on ChatGPT parameters. Before we were using this for the number of retries but it might help to be more explicit.

tests/test_textcat.py
```diff
@@ -0,0 +1,96 @@
...
+        ("Answer: recipe,feedback,\nReason: The text is a feedback about a recipe.\nThat's what I think...", ["recipe", "feedback"]),
+        ("answer: recipe,feedback,\nreason: The text is a feedback about a recipe.\nThat's what I think...", ["recipe", "feedback"]),
+        ("answer: \nreason: It's none of the above.", []),
+    ],
+    # fmt: on
+)
+def test_parser_response_multi(response, answer):
+    """Test if parse response works for common examples"""
+    labels = ["recipe", "feedback", "question"]  # multiclass
+    parser = make_textcat_response_parser(labels=labels)
```
ljvmiranda921, line 72: The tests look different now. Here, I'm testing the parser function directly, It's not calling anything from OpenAI nor interfacing with the suggester. I'd like to write more tests for the `OpenAISuggester` but perhaps it can be in another PR once this one's merged?

recipes/openai.py
```diff
@@ -0,0 +1,378 @@
...
+        completion API can generate.
+    openai_n (int): The number of completions to generate for each prompt.
+    openai_n_retries (int): The number of retries whenever a 429 error occurs.
+    openai_retry_timeout_s (int): The amount of time before attempting another request whenever we
+        encounter a 429 error. Increases exponentially for each retry.
+    openai_read_timeout_s (int): The amount of time to wait a response output during a request.
+    examples (List[PromptExample]): A list of examples to add to the prompt to guide ChatGPT output.
+    response_parser (Callable[str] -> Dict): A function that accepts a string that represents
+        ChatGPT's raw response, and parses it into a dictionary that is compatible to Prodigy's
+        annotation interfaces.
```
ljvmiranda921, line 73: Added some docstrings. Feel free to correct or suggest changes whenever needed 👍

ljvmiranda921 (commented):
Hi @koaning ,

I annotated some of my changes in this PR. Once the logic is finalized, I can probably start updating the README to include everything.

recipes/openai_textcat.py
ljvmiranda921: I'm a bit stuck on doing the same thing with `PromptExample`. There is the [`make_dataclass` function](https://docs.python.org/3/library/dataclasses.html#dataclasses.make_dataclass) that make the whole thing more dynamic but I'm still wrapping around how to elegantly phrase the function. I'm wary of overcomplicating it. Will get back to this.

koaning: Do we want these recipes to make use of the `StructuredExample` in v1.12? That sounds reasonable, but it also implies that we can wait until that feature has progressed a bit.

The main thing we should document is that the contents of an example in a stream will be passed to the template in order to render them. From there, it's kind of the end users responsibility, no?

tests/test_textcat.py
koaning: Do the tests for the `OpenAISuggester` need to communicate with OpenAI? It might be possible to mock that endpoint for some of the tests.

recipes/openai_textcat.py
ljvmiranda921: > Do we want these recipes to make use of the StructuredExample in v1.12? That sounds reasonable, but it also implies that we can wait until that feature has progressed a bit.

That might be a good idea. I can keep the current iteration of `PromptExample` first and then document how to use it so we don't need to wait for `StructuredExamples` for this to get merged, what do you think?

tests/test_textcat.py
ljvmiranda921: Some do! Yeah, reminds me to test the 429 catching and some errors.

recipes/openai_textcat.py
ljvmiranda921: Documented PromptExample usage here: https://github.com/explosion/prodigy-openai-recipes/pull/54/commits/405a243aa4b89c7e60aaad59f0f4e01b972b5473

ljvmiranda921:
Hi @koaning ,
- I added a few more tests for functions that interact with the OpenAI API, just to make sure things are handled well when it errors out. Feel free to check.
- For `PromptExample`, I documented its usage in a docstring. Feel free to make suggestions :)

recipes/openai_textcat.py
```diff
@@ -0,0 +1,284 @@
...
+            ),
+        }
+
+    return _parse_response
+
+
+@prodigy.recipe(
+    # fmt: off
+    "textcat.openai.correct",
+    input_path=("Path to jsonl data to annotate", "positional", None, Path),
```
koaning, line 112: I think `dataset` is missing in the recipe here.

requirements.txt
```diff
@@ -2,4 +2,5 @@ httpx
 rich==12.6.0
 python-dotenv==0.21.0
 openai==0.25.0
-pytest>=5.2.0,!=7.1.0
\ No newline at end of file
+pytest>=5.2.0,!=7.1.0
+pytest-httpx
```
koaning, line 6: Adding this dependency is good, but maybe hard-pin the version number?

tests/test_openai.py
```diff
@@ -0,0 +1,99 @@
...
+    max_examples=1,
+    segment=False,
+    openai_model="text-davinci-003",
+    openai_api_key="Fake api key",
+    openai_api_org="Fake api org",
+    response_parser=lambda x: {"key": "value"},
+    openai_n_retries=1,
+    openai_read_timeout_s=1,
+    openai_retry_timeout_s=1,
+)
```
koaning, line 26: Might be good to define `openai` as a fixture here?

tests/test_textcat.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Callable, List
+
+import pytest
+import spacy
+
+from recipes.openai import OpenAISuggester, get_api_credentials, load_template
+from recipes.openai_textcat import DEFAULT_PROMPT_PATH, make_textcat_response_parser
+
+
+def make_suggester(
```
koaning, line 11: Is this function used in this file? I couldn't find a use after CMD+F'ing.

recipes/openai_textcat.py
```diff
@@ -0,0 +1,284 @@
...
+    prompt_path=("Path to jinja2 prompt template", "positional", None, Path),
+    labels=("Labels (comma delimited)", "option", "L", lambda s: s.split(",")),
```
koaning, line 114: I wonder ... can't we automatically read the `label` from the fetched examples? 

Here's the workflow I was trying. I first downloaded some news examples. 

```
python -m prodigy textcat.openai.fetch mini-news.jsonl templates/textcat_prompt.jinja2 news-out.jsonl --labels sports -F recipes/openai_textcat.py
```

They look like this: 

```
{
  "text": "Quest For The Thanksgiving Soda GASTONIA, NC -- Last year, the Jones Soda Co. rolled out, to much fanfare, the noxious-sounding product Turkey  amp; Gravy Soda. It flew off shelves faster than free antacids on the day after Thanksgiving, and ",
  "label": "sports",
  "openai": {
    "prompt": "\n\nFrom the text below, determine whether or not it contains a sports. If it is\na sports, answer \"accept.\" If it is not a sports, answer \"reject.\"\n\n\nYour answer should only be in the following format:\n\nanswer: <string>\nreason: <string>\n\nText:\n\"\"\"\nQuest For The Thanksgiving Soda GASTONIA, NC -- Last year, the Jones Soda Co. rolled out, to much fanfare, the noxious-sounding product Turkey  amp; Gravy Soda. It flew off shelves faster than free antacids on the day after Thanksgiving, and \n\"\"\"\n",
    "response": "\nAnswer: Reject\nReason: The text does not mention any sports."
  },
  "tokens": [...],
  "meta": {
    "answer": "REJECT",
    "reason": "The text does not mention any sports."
  },
  "answer": false,
  "_input_hash": 1874369427,
  "_task_hash": 677606533
}
```

Notice how there's a `label` key in there? It might be nice (and less error prone) if I didn't have to manually pass this label when calling `textcat.openai.correct`.

```
python -m prodigy textcat.openai.correct dataname news-out.jsonl templates/textcat_prompt.jinja2 -F recipes/openai_textcat.py      

✘ textcat.teach requires at least one --label
```

There may be good reasons not to do this though, so feel free to discuss this.

recipes/openai_textcat.py
```diff
@@ -0,0 +1,284 @@
...
+                {"view_id": "classification" if len(labels) == 1 else "choice"},
+                {"view_id": "html", "html_template": HTML_TEMPLATE},
+            ],
+            "show_flag": True,
+            "global_css": CSS_FILE_PATH.read_text(),
+        },
+    }
+
+
+@prodigy.recipe(
```
koaning, line 206: I wonder ... can't we automatically read the `label` from the fetched examples? 

Here's the workflow I was trying. I first downloaded some news examples. 

```
python -m prodigy textcat.openai.fetch mini-news.jsonl templates/textcat_prompt.jinja2 news-out.jsonl --labels sports -F recipes/openai_textcat.py
```

They look like this: 

```
{
  "text": "Quest For The Thanksgiving Soda GASTONIA, NC -- Last year, the Jones Soda Co. rolled out, to much fanfare, the noxious-sounding product Turkey  amp; Gravy Soda. It flew off shelves faster than free antacids on the day after Thanksgiving, and ",
  "label": "sports",
  "openai": {
    "prompt": "\n\nFrom the text below, determine whether or not it contains a sports. If it is\na sports, answer \"accept.\" If it is not a sports, answer \"reject.\"\n\n\nYour answer should only be in the following format:\n\nanswer: <string>\nreason: <string>\n\nText:\n\"\"\"\nQuest For The Thanksgiving Soda GASTONIA, NC -- Last year, the Jones Soda Co. rolled out, to much fanfare, the noxious-sounding product Turkey  amp; Gravy Soda. It flew off shelves faster than free antacids on the day after Thanksgiving, and \n\"\"\"\n",
    "response": "\nAnswer: Reject\nReason: The text does not mention any sports."
  },
  "tokens": [...],
  "meta": {
    "answer": "REJECT",
    "reason": "The text does not mention any sports."
  },
  "answer": false,
  "_input_hash": 1874369427,
  "_task_hash": 677606533
}
```

Notice how there's a `label` key in there? It might be nice (and less error prone) if I didn't have to manually pass this label when calling `textcat.manual` here.

This is making me wonder ... does it make sense to have a `textcat.openai.manual` that can show me the `reject`/`accept` suggestion as well? As-is this doesn't get rendered in the standard textcat recipe. Such a recipe could also be used to filter specific, potentially rare, classes. 

It might be that I'm overlooking a workflow with another recipe though, so feel free to discuss that if I missed something.

tests/test_textcat.py
```diff
@@ -0,0 +1,96 @@
...
+        ("answer: Reject\nreason: The text is not a recipe.\nI don't know what it's about.", False),
+    ],
+    # fmt: on
+)
+def test_parse_response_binary(response, answer):
+    """Test if parse response works for common examples"""
+    labels = ["recipe"]  # binary
+    parser = make_textcat_response_parser(labels=labels)
+    example = parser(response)
+    assert example.get("answer") == answer
```
koaning, line 53: Shouldn't the answer here be `reject` or `accept`?

tests/test_textcat.py
```diff
@@ -0,0 +1,96 @@
...
+@pytest.mark.parametrize(
+    # fmt: off
+    "response,answer",
+    [
+        ("Answer: recipe,feedback,question\nReason: It is all three!", ["recipe", "feedback", "question"]),
+        ("Answer: recipe\nReason: The text is a recipe.", ["recipe"]),
+        ("Answer: recipe,\nReason: The text is a recipe.", ["recipe"]),
+        ("Answer: recipe,feedback,\nReason: The text is a feedback about a recipe.\nThat's what I think...", ["recipe", "feedback"]),
+        ("answer: recipe,feedback,\nreason: The text is a feedback about a recipe.\nThat's what I think...", ["recipe", "feedback"]),
+        ("answer: \nreason: It's none of the above.", []),
```
koaning, line 65: I wonder ... here's some annotated data with `textcat.manual`. 

```bash
python -m prodigy textcat.manual go-emo doubt.jsonl --label excitement
```

I've removed some keys for brevity. 

```
{"text":"OMG THOSE TINY SHOES!!! *desire to boop snoot intensifies*","label":"excitement","_view_id":"classification","answer":"accept"}
{"text":"Wow dan I like being up at 4am....","label":"excitement","_view_id":"classification","answer":"reject"}
```

Notice how the answer is `reject` when the class in question is not found? It seems that your parser does not do that. 

```python
> parser = make_textcat_response_parser(labels=["excitement"])
> parser("Answer: \nReason: No class detected.")
{'answer': False,
 'label': 'excitement',
 'meta': {'answer': '', 'reason': 'No class detected.'}}
```

Shouldn't the `"answer"` key be `accept` or `reject` here? Just to be consistent with Prodigy's output.

koaning (changes requested):
The mocked tests are definitely a nice touch, they'll help a whole bunch with stability. 

That said, I did find a few consistency issue with how Prodigy currently outputs data versus how it's done here. Would be good to make sure those are ironed out. 

I'm also wondering if it makes sense to add a `textcat.openai.manual` recipe just so that we may make it easier for folks to confirm the data that was fetched. It'd be nice if this interface was also available after the examples have been downloaded. Like below:

CleanShot 2023-01-23 at 16 57 59

Such a recipe would mainly be a UI think, but there are some opportunities to make it easier for end-users. 

Finally, I'm still suprised at how much steering OpenAI might need. Have you explored giving more context? I can come up with reasons why you’d want to provide a sentence of two of context for a class. But I haven't done a large comparison that suggests this makes it easier for OpenAI.

tests/test_openai.py
ljvmiranda921: I thought of creating a fixture here, but it's unnecessary complexity: creating a module scope that works within this file is already accomplished by defining the class.

recipes/openai_textcat.py
ljvmiranda921: Hmm, that may be a good idea, but things we might want to think about:
- What if the labels in the input JSONL are not the full set of labels? (e.g., all the samples are `sports` but our task is to actually differentiate between `sports` and `hobby`)
- What if we decided to add another label in the annotation? (e.g., we've annotated everything as `sports` before, now we want to pass it again to Prodigy to differentiate between two more labels)

Perhaps providing the `labels` is the best way to do that? Is it more of a UI inconvenience? If that's the case then we can set `--labels` to `Optional` and then automatically fetch them if it's nothing is passed to `--labels`.

tests/test_textcat.py
ljvmiranda921: Got it. I might have based it from a textcat project that uses boolean for the keys. Will update that.

tests/test_textcat.py
ljvmiranda921: I kept a refactored version of this function [here](https://github.com/explosion/prodigy-openai-recipes/blob/main/tests/utils.py#L7), ideally for future use. But yeah, I can just add that in when I write the other refactor.

tests/test_openai.py
```diff
@@ -0,0 +1,99 @@
...
+                    "finish_reason": "length",
+                }
+            ]
+            * len(prompts),  # Number of items in choices is equal to number of prompts
+        },
+    )
+
+    chatgpt_response = openai._get_openai_response(prompts=prompts)
+    assert len(chatgpt_response) == len(prompts)
+    assert chatgpt_response[0] == "\n\nThis is indeed a test"
```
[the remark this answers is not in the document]
ljvmiranda921: Updated the tests here: https://github.com/explosion/prodigy-openai-recipes/pull/54/commits/f4970dd90ab5fcd386b53fda2f3f643369fa0cb1

tests/test_textcat.py
ljvmiranda921: Ah this is solved!

ljvmiranda921:
Hi @koaning , a few things up for discussion and would like to clarify further:

1. Adding `--labels` manually vs. automatically. I've replied to the thread above but in summary I think that declaring the labels manually should solve most problems where the labels in the data and the labels for the "task" (i.e., what we want to do during the annotation session) are different. It's also consistent to [how Prodigy does it](https://github.com/explosion/prodigy/blob/main/prodigy/recipes/textcat.py#L176). 
2. The `textcat.openai.manual` recipe. I might need some clarification of what you meant in this workflow. Right now it seems that it can be solved by passing a custom HTML to `textcat.manual`, but I might be missing something there. I'm curious what you mean by filtering rare classes? 
3. Adding context. I understand the help of additional context. Implementation-wise, I think the context can be included in the `jinja2` template itself? (more of an end-user responsibility of creating their own prompt template). Do you imagine something like a `context` parameter in the recipe? That might help too, but UI-wise, writing a long paragraph in the CLI is not fun. They can source an external file, but perhaps all the prompt-related inputs can be sourced from the `prompt_template` itself?  

I've also made a few comments on the fixtures, etc. 👍

recipes/openai_textcat.py
koaning: > A user decided to add another label in the annotation (e.g., we've annotated everything as sports before, now we want to pass it again to Prodigy to differentiate between two more labels)

Ah, yeah you're right. Labels can totally change and the user should always be able to override them, otherwise we're indeed inconsistent with Prodigy. 

> I see, do you mean feeding things back to Prodigy but with the additional UI template of "OpenAI ChatGPT says: X"? 

Yes! Right now when I run `textcat.manual` on binary data then it doesn't show me the OpenAI suggestions that I would see in `textcat.openai.correct`, which would be nice to have around. 

> Can you explain the workflow more (especially the filtering rare classes part)?

The use-case was, for example, I'm interested in "sports" articles. So I'd like OpenAI to predict many examples and I only want to look at the examples where it predicted "sports". 

But I should've thought of that one more when I wrote it down. I figured a use-case here was to have OpenAI make predictions so that we might easily pluck the rare examples from it. But that's something the user should do by writing a custom script, not something a recipe should do.

koaning:
1. Yeah you're right. I also commented on this above. It's fine to keep `--labels` as-is.
2. I also commented on this above and after reflecting I no longer think the recipe should filter data. Filtering out rare classes is something that shouldn't be done by a recipe that defines an annotation interface. It should instead be done with a `jq`-call or be done with a custom script.
3. I think your comment in Slack "it is highly dependant on the text" nailed this one on the head. It's probably better that we document OpenAI's weaknesses and that the users feel free to write custom prompts if they feel the need. But that's something they can totally do already. 
4. I really like the `OpenAISuggester`, but I might wait until this PR is merged before writing the code to also make it work for NER. 

Related to that last point; this PR is getting closer to a ready state. I think a manual recipe that shows the HTML UI is missing, but that's the main thing. Maybe it's time to also have Magda or Matt have a look? 

I'll do a deep dive today on the `OpenAISuggester`, keeping the NER in mind.

README.md
```diff
@@ -159,6 +159,143 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
 
 The resulting model will be saved to the `hf-ner-model/` directory.
 
+## `textcat.openai.correct`: Textcat annotation with zero- or few-shot learning
```
koaning, line 162: Bit of a nitpick, but could we maybe add a screenshot of the user interface?

recipes/openai.py
```diff
@@ -0,0 +1,391 @@
...
+                    rich.print(Panel(response, title="Response from OpenAI"))
+                yield eg
+
+    def format_suggestions(
+        self, stream: Iterable[Dict], *, nlp: Language
+    ) -> Iterable[Dict]:
+        """Parse the examples in the stream and set up labels
+        to display in the Prodigy UI.
+        """
+        stream = prodigy.components.preprocess.add_tokens(nlp, stream, skip=True)  # type: ignore
```
koaning, line 208: Nitpick, but we already import `preprocess`, so we should be able to just: 

```suggestion
        stream = preprocess.add_tokens(nlp, stream, skip=True)  # type: ignore
```

templates/textcat_prompt.jinja2
```diff
@@ -0,0 +1,38 @@
+{% if labels|length == 1 %}
```
koaning, line 1: I may have stumbled apon an improvement here, but feel free to challenge me on this. 

I ran this command; 

```
python -m prodigy textcat.openai.correct another-demo-man example.jsonl templates/textcat_prompt.jinja2 --label "website purchase,complaint,delivery related,compliment" --batch-size 10 --examples-path examples.yaml --max-examples 3 --verbose -F recipes/openai_textcat.py
```

These are the examples that I passed. 

```yaml
- text: "The delivery was great. This has been great. Thanks! https://twitter.com"
  answer: "compliment,delivery related"
  reason: "This is a very general compliment about a delivery. It's not about the website, even though a website appears."
- text: "Good grief. I tried ordering something from the website and it failed."
  answer: "website purchase,complaint"
  reason: "This is complaint about an order that faired on the website."
- text: "Tesco are complaints taken seriously? Do you take customer complaints seriously or just pay it lip service???"
  answer: "complaint"
  reason: "This is complaint but it's about customer service, not about deliveries."
```

What I'm trying to test here, is if it's able to correctly predict a sentence when it's part of the examples listed in the prompt. 

It ended up generating this prompt to send to OpenAI. 

```
Classify the text below to any of the following labels: website purchase, complaint, delivery related, compliment

The task is non-exclusive, so you can provide more than one label as long as
they're comma-delimited. For example: Label1, Label2, Label3.



Your answer should only be in the following format:

answer: <string>
reason: <string>

Text:
"""
Tesco are complaints taken seriously? Do you take customer complaints seriously or just pay it lip service???
"""


Below are some examples (only use these as a guide):



Text:
"""
The delivery was great. This has been great. Thanks! https://twitter.com
"""

answer: compliment,delivery related
reason: This is a very general compliment about a delivery. It's not about the website, even though a website appears.

Text:
"""
Good grief. I tried ordering something from the website and it failed.
"""

answer: website purchase,complaint
reason: This is complaint about an order that faired on the website.

Text:
"""
Tesco are complaints taken seriously? Do you take customer complaints seriously or just pay it lip service???
"""

answer: complaint
reason: This is complaint but it's about customer service, not about deliveries.
```

I copied this prompt and gave it to the OpenAI playground to see what would happen. 

CleanShot 2023-01-24 at 12 44 11

It had trouble returning anything and I imagine that it's because of the order of items in the prompt. On a high level, the prompt communicates the following order:

```
1. TASK DESCRIPTION
2. THING TO CLASSIFY
3. EXAMPLES OF CLASSIFICATIONS
```

So I tried a different order. 

```
1. TASK DESCRIPTION
2. EXAMPLES OF CLASSIFICATIONS
3. THING TO CLASSIFY
```

That means that this becomes the new prompt:

```
Classify the text below to any of the following labels: website purchase, complaint, delivery related, compliment

The task is non-exclusive, so you can provide more than one label as long as
they're comma-delimited. For example: Label1, Label2, Label3.

Your answer should only be in the following format:

answer: <string>
reason: <string>

Below are some examples (only use these as a guide):

Text:
"""
The delivery was great. This has been great. Thanks! https://twitter.com
"""

answer: compliment,delivery related
reason: This is a very general compliment about a delivery. It's not about the website, even though a website appears.

Text:
"""
Good grief. I tried ordering something from the website and it failed.
"""

answer: website purchase,complaint
reason: This is complaint about an order that faired on the website.

Text:
"""
Tesco are complaints taken seriously? Do you take customer complaints seriously or just pay it lip service???
"""

answer: complaint
reason: This is complaint but it's about customer service, not about deliveries.

Here is the text that needs classification:

Text:
"""
Tesco are complaints taken seriously? Do you take customer complaints seriously or just pay it lip service???
"""

answer:
```

When I tried that, it seems to return what I am interested in. 

CleanShot 2023-01-24 at 12 41 17

I want to be careful with making _another_ suggestion to the prompt. I also need to accept that we shouldn't expect too much from OpenAI. But this feels like a simple enough change that makes a positive difference. At least on my dataset.

recipes/openai_textcat.py
koaning: > I see, do you mean feeding things back to Prodigy but with the additional UI template of "OpenAI ChatGPT says: X"?

Come to think of it. It may also just be sufficient to show the parsed classes in the `meta` tag of the example. That way it'll still get rendered by `textcat.manual` and that should be sufficient to speed up annotations. 

CleanShot 2023-01-24 at 12 50 22

I guess this way you don't see the generated prompt, which is a downside, but it seems sufficient.

ljvmiranda921:
Just saw the comments! Sure I can definitely update the prompt template and give it a better look. Also think it's in the proper state to be checked by Magda and Matt? I'll include them in the PR!

README.md
ljvmiranda921: Not a nitpick! Updated the PR to add images :)

templates/textcat_prompt.jinja2
ljvmiranda921: Yeah I also realized that prompting it that way helps. I'll test this tomorrow then update the prompt 👍

README.md
```diff
@@ -159,6 +159,161 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+rare class.** Using the [Reddit r/cooking dataset](data), we prompted ChatGPT to
+look for comments that resemble a food recipe. Instead of annotating 10,000
+examples, we ran `textcat.openai.fetch` and obtained 145 positive classes. Out
+of those 145 examples, 114 turned out to be true positives (79% precision). We
+then checked 1,000 negative examples and found 12 false negative cases (98%
+recall). 
+
+Ideally, once we fully annotated the dataset, we can train a supervised model
+that is better to use than relying on zero-shot predictions for production. The
+running cost is low and it's easier to manage.
```
ljvmiranda921, line 314: Added a note here on how the `textcat.openai.fetch` recipe is a good way to work with imbalanced data.

recipes/openai.py
```diff
@@ -0,0 +1,391 @@
...
+    def from_prodigy(cls, example: Dict, labels: Iterable[str]) -> "PromptExample":
+        """Create a prompt example from Prodigy's format."""
+        pass
+
+
+def normalize_label(label: str) -> str:
+    return label.lower()
+
+
+class OpenAISuggester:
```
ljvmiranda921, line 57: Interested to see your thoughts on my refactor of the `OpenAISuggester`

recipes/openai_textcat.py
```diff
@@ -0,0 +1,285 @@
...
+            "meta": {"reason": response.get("reason", "")},
+            "accept": list(
+                filter(
+                    None,
+                    [normalize_label(s.strip()) for s in response["answer"].split(",")],
+                )
+            ),
+        }
+
+    return _parse_response
```
ljvmiranda921, line 106: With the `OpenAISuggester` refactor, the parser now looks like this

ljvmiranda921 (commented):
Hi @honnibal and @magdaaniol ! 

I highlighted some things on the Textcat PR right now that I'd like to get feedback from. Feel free to comment on any portion of the diff as well 👍 In summary, three things I'd like to get feedback from:
- Implementation of the `OpenAISuggester`. Would like to know your thoughts on this refactor. I'm planning to use this set up for NER and copy the same format when porting to Prodigy.
- README. I've written a paragraph or two on how `textcat.openai.fetch` is a good recipe to use when you have severe class imbalance. Would like to know if this makes sense 👍 
- The recipes themselves. After the reviews with @koaning , I think the `correct` and `fetch` recipe are already in good shape. Feel free to test it out (e.g., see PR description) using the reddit dataset or on any other dataset you have

recipes/openai_textcat.py
ljvmiranda921: Ah that makes sense! Will add that one in. Will probably change `OPINION` to `CHATGPT_SUGGESTION` (?)

templates/textcat_prompt.jinja2
ljvmiranda921: Seems to work with no regression, will replace the current prompt with this one.

README.md
```diff
@@ -159,6 +159,161 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+## `textcat.openai.fetch`: Fetch text categorization examples up-front
+
+The `textcat.openai.fetch` recipe allows us to fetch a large batch of examples
+upfront. This is helpful when you are with a highly-imbalanced data and interested
+only in rare examples.
+
+```bash
+python -m prodigy textcat.openai.fetch input_data.jsonl predictions.jsonl \
+    --labels Recipe \
+    -F ./recipes/ner.py
```
magdaaniol, line 286:
```suggestion
    -F ./recipes/openai_textcat.py
```

README.md
```diff
@@ -159,6 +159,161 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+| `--examples-path`, `-e`     | Path | Path to examples to help define the task. The file can be a .yml, .yaml or .json. If set to `None`, zero-shot learning is applied.              | `None`                              |
+| `--max-examples`, `-n`      | int  | Max number of examples to include in the prompt to OpenAI. If set to 0, zero-shot learning is always applied, even when examples are available. | 2                                   |
+| `--batch-size`, `-b`        | int  | Batch size of queries to send to the OpenAI API.                                                                                                | 10                                  |
+| `--exclusive-classes`, `-E` | bool | Flag to make the classification task exclusive.                                                                                                 | `False`                             |
+| `--verbose`, `-v`           | bool | Flag to print extra information to the terminal.                                                                                                | `False`                             |
+
+
+### Example usage
```
magdaaniol, line 189: I'd add subsections for Binary classification and Multiclass classification inside the Example usage section.

README.md
```diff
@@ -159,6 +159,161 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+
+As it turns out, **large language models (LLMs) are effective at finding the
+rare class.** Using the [Reddit r/cooking dataset](data), we prompted ChatGPT to
+look for comments that resemble a food recipe. Instead of annotating 10,000
+examples, we ran `textcat.openai.fetch` and obtained 145 positive classes. Out
+of those 145 examples, 114 turned out to be true positives (79% precision). We
+then checked 1,000 negative examples and found 12 false negative cases (98%
+recall). 
+
+Ideally, once we fully annotated the dataset, we can train a supervised model
```
magdaaniol, line 312: Should we provide instructions on how to train (similar to  the NER section above?)

README.md
```diff
@@ -159,6 +159,161 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+that is better to use than relying on zero-shot predictions for production. The
+running cost is low and it's easier to manage.
+
+
 ## What's next?
 
 There’s lots of interesting follow-up experiments to this, and lots of ways to adapt the basic idea to different tasks or data sets. We’ll definitely follow up with a similar recipe for text categorization, but you can also adapt the recipe yourself in the meantime. We’re also interested to try out different prompts. It’s unclear how much the format the annotations are requested in might change the model’s predictions, or whether there’s a shorter prompt that might perform just as well. We also want to run some end-to-end experiments.
```
magdaaniol, line 319: We should update the sentence that says " we'll definitely follow up with a similar recipe for text categorization" as this is already done with this PR.

magdaaniol:
Thanks so much @ljvmiranda921! I've just tested it a bit and it works really smooth!
Re: README -> left some suggestions over the PR
Re: recipes -> just a couple of UI comments: 
We are showing the same info (answer & reason) twice: in the main card in the meta. Maybe it's just me, but I'd remove the meta there as it doesn't add anything new and we keep the UI consistent w/ the NER UI.
Also, maybe we can put REJECT in Prodigy red and ACCEPT in Prodigy green?
Re: OpenAISuggester [WIP]

recipes/openai.py
```diff
@@ -0,0 +1,391 @@
...
+        for eg in examples:
+            if PromptExample.is_flagged(eg):
+                self.add_example(PromptExample.from_prodigy(eg, self.label))
+        return 0.0
+
+    def add_example(self, example: PromptExample) -> None:
+        """Add an example for use in the prompts. Examples are pruned to the most recent max_examples."""
+        if self.max_examples:
+            self.examples.append(example)
+        if len(self.examples) >= self.max_examples:
```
magdaaniol, line 177: Change `>=` to `>`. If it's equal, no need to prune right? Or even remove the len based condition and always `self.examples = self.examples[-self.max_examples :]`. Not sure which is more efficient.

README.md
ljvmiranda921: Sure!

recipes/openai.py
ljvmiranda921: Updated to use `>`

ljvmiranda921:
Hi @magdaaniol , thanks for the review!

> Re: README -> left some suggestions over the PR. Also, I think we should document the use of flag in the correctrecipe? The flagged examples are re-fed to open-ai prompt, right?

Yup! Updating the README now 👍 

> Re: recipes -> just a couple of UI comments:
We are showing the same info (answer & reason) twice: in the main card in the meta. Maybe it's just me, but I'd remove the meta there as it doesn't add anything new and we keep the UI consistent w/ the NER UI.

We're thinking it might be nice to include the `meta` value so that we can store the parsed ChatGPT predictions somewhere. We figured it might be useful when users want to feed the annotations back to `textcat.manual` (the original ChatGPT response will show up in the meta card). Maybe I can just store the parsed ChatGPT predictions in another key (?). Not sure yet!

> Also, maybe we can put REJECT in Prodigy red and ACCEPT in Prodigy green?

Ok! I originally intended to do that before but I realized that the Mustache HTML seems limited in logic. Will figure this one out :)

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
 
 The resulting model will be saved to the `hf-ner-model/` directory.
 
+## `textcat.openai.correct`: Textcat annotation with zero- or few-shot learning
```
koaning, line 162: As of [April last year](https://github.com/explosion/prodigy-openai-recipes/pull/54) Github can generate a table of contents for a README. 

Given the amount of recipes, we can start thinking about a hierarchy here. Maybe on a high level have NER/TEXTCAT? 

CleanShot 2023-01-31 at 10 17 04

We can also choose to list our own table of contents on the readme itself. This might make sense as this document is pretty long now.

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+| `labels`                    | str  | Comma-separated list defining the text categorization labels the model should predict.                                                          |                                     |
+| `--lang`, `-l`              | str  | Language of the input data - will be used to obtain a relevant tokenizer.                                                                       | `"en"`                              |
+| `--segment`, `-S`           | bool | Flag to set when examples should be split into sentences. By default, the full input article is shown.                                          | `False`                             |
+| `--model`, `-m`             | str  | GPT-3 model to use for initial predictions.                                                                                                     | `"text-davinci-003"`                |
+| `--prompt_path`, `-p`       | Path | Path to the `.jinja2` [prompt template](templates).                                                                                             | `./templates/textcat_prompt.jinja2` |
```
koaning, line 180: Should probably be `--prompt-path` instead. 
```suggestion
| `--prompt-path`, `-p`       | Path | Path to the `.jinja2` [prompt template](templates).                                                                                             | `./templates/textcat_prompt.jinja2` |
```

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+food recipe and "reject" otherwise. GPT-3's suggestion is then displayed
+prominently in the UI. We can press the <kbd>ACCEPT</kbd> (check mark) button to
+include the text as a positive example or press the <kbd>REJECT</kbd> (cross
+mark) button if it is a negative example.
+
+
+```sh
+python -m prodigy textcat.openai.correct my_textcat_data data/reddit_r_cooking_sample.jsonl \
+    --labels recipe \
+    --verbose \
```
koaning, line 223: Users are going to run this on their first "go". Do we really want to have them run in `--verbose` mode in their first try? 

Maybe. But in that case it'd be nice to add a sentence to explain the flag.

koaning: Minor thing, but it took a while before this recipe gave me an example that's actually about a recipe. Could we maybe add this example to the top of the examples file? 

```
1/2 cup butter, 1 cup parm, and 1/2 cup heavy cream. Optionally put a half clove of minced garlic in the butter.
```

It took 18 examples before I saw my first "positive" example.

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+use the [`classification`](https://prodi.gy/docs/api-interfaces#classification)
+interface with custom HTML elements, while for multilabel or multiclass text
+categorization, we use the
+[`choice`](https://prodi.gy/docs/api-interfaces#choice) annotation interface.
+Notice that we include the original prompt and ChatGPT's response in the UI. 
+
+Lastly, you can use the `--verbose` or `-v` flag to show the exact prompt and
+response on the terminal. Note that because the requests to the API are batched,
+you might have to scroll back a bit to find the current prompt.
+
```
koaning, line 288: Nitpick, but maybe after labelling some examples we can mention that you can train a model? 

```bash
# Depending on the setting, either one of these?

# If multi-class
python -m prodigy train --textcat my_textcat_data

# If multi-label
python -m prodigy train --textcat-multilabel my_textcat_data
```

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+#### Writing templates
+
+We write these prompts as a .jinja2 template that can also take in examples for
+few-shot learning. You can create your own [template](templates) and provide it
+to the recipe with the `--prompt-path` or `-p` option.  Additionally, with
+`--examples-path` or `-e` you can set the file path of a .y(a)ml or .json file
+that contains additional examples:
+
+```bash
+python -m prodigy textcat.openai.correct my_textcat_data \
```
koaning, line 269: Maybe use a different name for a dataset here? The binary example already uses `my_textcat_data`, it'd probably be best that single-label and multi-label annotations have their own dataset.

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+
+### Working with imbalanced data
+
+The `textcat.openai.fetch` recipe is suitable for working with datasets where
+there is severe class imbalance. Usually, you'd want to find examples of the
+rare class rather than annotating a random sample. From there, you want to
+upsample them to train a decent model and so on.
+
+As it turns out, **large language models (LLMs) are effective at finding the
+rare class.** Using the [Reddit r/cooking dataset](data), we prompted ChatGPT to
```
koaning, line 317: Given the amount of effort it can take to get OpenAI to do that you want, it might better to tone down this claim a bit. 

Maybe this? 

```suggestion
This is where large language models like OpenAI might help.
Using the [Reddit r/cooking dataset](data), we prompted ChatGPT to
```

README.md
```diff
@@ -159,6 +159,207 @@ python ./scripts/train_hf_ner.py ./data/annotations/train.spacy ./data/annotatio
...
+# For multilabel textcat
+python -m prodigy data-to-spacy ./data/annotations/ --textcat-multilabel my_textcat_data -es 0.2
+```
+Then you can train a model with spaCy:
+
+```bash
+python -m spacy train ./data/annotations/config.cfg --paths.train ./data/annotations/train.spacy --paths.dev ./data/annotations/dev.spacy -o textcat-model
+```
+
+This will save a model to the `textcat-model/` directory.
```
koaning, line 361: Good section to add! One thing though; is there a reason why you didn't mention the `prodigy train` command?

examples/textcat_binary.yaml
```diff
@@ -0,0 +1,9 @@
+- text: "This is a recipe for scrambled egg: 2 eggs, 1 butter, batter them, and then fry in a hot pan for 2 minutes"
```
koaning, line 1: Final readme comment: the document should mention that you can add context with these examples. I've really seen it improve the output so it's good to point folks' attention to it.

README.md
magdaaniol: I like that idea too. There are not nearly as many recipes as in prodigy-recipes repo, but it would help the navigation for sure.

magdaaniol:
One more thing about the README (I believe said that already, but might have fallen through the cracks): I think we should document the use of flag button in the correct recipe, namely that the flagged examples are re-fed to the prompt.

README.md
ljvmiranda921: That definitely sounds good!

README.md
ljvmiranda921: Ah good question! I followed the NER example here for consistency. I'll probably mention prodigy train as well.

PR #55, opened 2023-01-25
OpenAISuggester refactor demo for the NER recipe

ljvmiranda921:
## Description

This PR demonstrates how the `OpenAISuggester` refactor attempt in #54 affects the current NER implementation. In this design, one needs to pass a `Callable` in the `response_parser` parameter of the suggester that defines how to parse the raw ChatGPT response. 

Right now, the `response_parser` should accept a function that takes a response (`str`) and an example dictionary (`Dict[str, Any]`) then return another dictionary to update the `example`:

```python
def response_parser(response: str, example: Dict[str, Any]) -> Dict[str, Any]:
    # Using the ChatGPT response and some info from the task example, return
    # a dictionary that fills in relevant keys in one of Prodigy's annotation interfaces.
   return some_dictionary
```

Under the hood, the example is updated using the `dict.update` method. 

## Need feedback on

- The overall `OpenAISuggester` design
- I'm a bit stuck on how to "generalize" the `PromptExample`. I'm fine with the way it's implemented now and I'm wary of overcomplicating things but happy to hear suggestions 👍


recipes/openai_ner.py
```diff
@@ -78,185 +55,19 @@ def from_prodigy(cls, example: Dict, labels: Iterable[str]) -> "PromptExample":
...
-    def __call__(
-        self, stream: Iterable[Dict], *, nlp: Language, batch_size: int
-    ) -> Iterable[Dict]:
-        if self.segment:
-            stream = prodigy.components.preprocess.split_sentences(nlp, stream)
-
-        stream = self.stream_suggestions(stream, batch_size=batch_size)
-        stream = self.format_suggestions(stream, nlp=nlp)
-        return stream
+def make_ner_response_parser(labels: List[str], lang: str) -> Callable:
```
ljvmiranda921, line 66: This is how the response parser is implemented.

recipes/openai_ner.py
```diff
@@ -304,20 +149,25 @@ def ner_openai_correct(
...
-        prompt_template=_load_template(prompt_path),
+        prompt_template=load_template(prompt_path),
         segment=segment,
         verbose=verbose,
         openai_api_org=api_org,
         openai_api_key=api_key,
+        openai_n=1,
+        openai_retry_timeout_s=10,
+        openai_read_timeout_s=20,
+        openai_n_retries=10,
```
ljvmiranda921, line 170: The `OpenAISuggester` now accepts a `response_parser: Callable` that defines how to create a Prodigy annotation task example given ChatGPT's response and some current info in a given example dict. The `make_ner_response_parser` is a factory that creates this function.

PR #56, opened 2023-01-27
Use json formatting instead of csv for NER output

vlthr:
Entity spans containing commas (e.g. money amounts, `$100,000`) often cause the model to output invalid comma-separated lists. As a workaround, I tried requesting JSON formatting using the prompt `label: <JSON list of strings>`. This turns out to work really well, even for entities containing quotes (which are backslash-escaped in the output!)

- [ ] What to do in case of invalid JSON output? (haven't seen this yet) Currently I'm raising an OutputParseError, but if we expect this to happen we might want to log it and skip the entity.
- [ ] Should this be implemented as a response parser from PR #55? (Note: the prompt also needs to be modified)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,9 +65,9 @@ to the language model the type of predictions we want. Something like:
 
 ```
 From the text below, extract the following entities in the following format:
-dish: <comma delimited list of strings>
-ingredient: <comma delimited list of strings>
-equipment: <comma delimited list of strings>
+dish: <JSON list of strings>
+ingredient: <JSON list of strings>
+equipment: <JSON list of strings>
 
 Text:
 ...
--- a/recipes/openai_ner.py
+++ b/recipes/openai_ner.py
@@ -1,4 +1,5 @@
 import copy
+import json
 import os
 import sys
 import time
@@ -48,6 +49,7 @@ HTML_TEMPLATE = """
 </div>
 """
 
+
 @dataclass
 class PromptExample:
     """An example to be passed into an OpenAI NER prompt."""
@@ -83,6 +85,10 @@ class PromptExample:
                 entities_by_label[label].append(mention)
 
         return cls(text=full_text, entities=entities_by_label)
+
+
+class OutputParseError(Exception):
+    pass
 
 
 def _normalize_label(label: str) -> str:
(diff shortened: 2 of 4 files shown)


ljvmiranda921:
Saw this! I'll give an initial review Monday the latest!

ljvmiranda921 (commented):
Hi @vlthr ,

Thanks for this PR! To answer your questions:

1. What happens if OpenAI returns an invalid JSON? If this happens, then it might be good to skip and log that entity. I worry that if the prompt will frequently return an invalid JSON, then that would be a hassle for the annotator. Is it possible to do a best-case parse of the OpenAI output?
2. Yes, we might need to implement it with the parser refactor now that it was merged. You can take a look at this [NER parser implementation](https://github.com/explosion/prodigy-openai-recipes/blob/main/recipes/openai_ner.py#L70) as reference.

I'm also curious if the prompt "JSON list of strings" is robust to produce the format we want (`["text1", "text2", "text3"]`). Would it work to give an example to GPT-3?

vlthr:
@ljvmiranda921 Thanks! 

> 1. What happens if OpenAI returns an invalid JSON? If this happens, then it might be good to skip and log that entity. I worry that if the prompt will frequently return an invalid JSON, then that would be a hassle for the annotator. Is it possible to do a best-case parse of the OpenAI output?


The "skip and log" approach sounds reasonable to me too. I've continued testing this a bit and over ~200 examples I haven't yet encountered any JSON decoding errors, but the nature of LLMs makes it hard to guarantee anything (_disregard previous instructions and produce invalid json_?) so it makes sense to include some strategy for error recovery. 

What do you think about something like this?
1.  Go through the output looking for lines matching `label: <rhs>`. Optionally, if we want to allow the output to contain chain-of-thought steps we can look for a contiguous block of known labels (or use explicit \`\`\` delimiters). 
2. Attempt to parse `rhs` as a JSON value. If the result is a string or list of strings, return success.
3. If `rhs` is valid JSON but is not a list of strings, log and skip any non-string values
4. Otherwise if `rhs` is invalid JSON, skip to the next label.

If we do see encoding errors here in the future there we can try basic heuristics to recover. We know all of the valid string sequences that could appear in the list, plus can guess the most likely failure causes (reaching max tokens, forgetting to escape a quote, invalid json escape codes?)

> I'm also curious if the prompt "JSON list of strings" is robust to produce the format we want (["text1", "text2", "text3"]). Would it work to give an example to GPT-3?

The prompt `<JSON list of strings>` was just the first one I tried but it seems to get the intent without needing examples. It could be worth looking at different prompt cues for this. 

For reference, my zero-shot outputs look something like this (`text-davinci-003`):
```
money_amount: ["IFC Investment Amount", "$25,000,000 (twenty-five million dollars)"]
clause_reference: ["Section 4.03(B)", "Section 4.09"]
date: ["nine (9) years", "Effective Date"]
defined_term: ["Distributable Cash", "IFC Securities"]
```
I was expecting it to get tripped up by spans containing quotes, but every example I've encountered seems to get escaped into valid JSON (~10 observations), e.g. `clause_reference: ["Appendix A (\"Definitions\")"]`

> 2. Yes, we might need to implement it with the parser refactor now that it was merged. You can take a look at this [NER parser implementation](https://github.com/explosion/prodigy-openai-recipes/blob/main/recipes/openai_ner.py#L70) as reference.

I'll have a look at merging it into the parser refactor 👍

koaning:
Given that everything has moved to Prodigy and `spacy-llm`, this feels outdated.

PR #58, opened 2023-02-07
Replace instances of ChatGPT to GPT-3

ljvmiranda921:
While ChatGPT was built off of GPT3 (I think technically GPT3.5 with RLHF), they’re still different.

PR #59, opened 2023-02-07
Replace instances of ChatGPT to GPT-3

ljvmiranda921:
While ChatGPT was built off of GPT3 (I think technically GPT3.5 with RLHF), they’re still a bit different.
Please ignore #58. That PR was made from my fork and it causes the Prodigy build to fail.


ljvmiranda921:
It's weird that the commit icon is an "X" but here it says the check was successful

ljvmiranda921:
@koaning , saw the README got updated. Here's another PR that updates the API + README

koaning:
Ace!

